PR #5, opened 2025-11-17
Fix README flags table formatting

mrshu:
It was a bit broken previously. The escaping might help.


mrshu:
Thanks a bunch for oracle @steipete! I've noticed the README being a bit broken while reading through it -- the escaping here should help.

Thanks again!

steipete:
Thanks!

PR #9, opened 2025-11-17
fix: Improve model selection scoring

christomitov:
GPT5 Pro was not being selected on enterprise accounts and was defaulting to GPT5, this addresses that.

Before:

image

After:

image


steipete:
Thank you!

PR #13, opened 2025-11-18
feat: add configurable base URL 

Tr1Fecta-7:
## Why
Until now the CLI always talked to api.openai.com with the openai package. Anyone routing through LiteLLM, Azure OpenAI, or an internal proxy had to rewrite the binary. This change makes the endpoint first-class so they can point Oracle at whatever OpenAI-compatible gateway they run. 

Vibed this with GPT 5.1 Codex + some manual changes, if I should make any changes please let me know! I would love to see this get merged in.  

## What’s new
- `--base-url <url>` flag on the CLI. Example:
```
  OPENAI_API_KEY=sk-*** oracle \
    --engine api \
    --base-url https://gateway.example.com/v1 \
    --prompt "Summarize the changelog" \
    --wait --verbose
```
    
  Sample output now includes:
    `Using OPENAI_API_KEY=sk-A***Z`
    `Using OpenAI base URL https://gateway.example.com/v1`
- `~/.oracle/config.json` gains `apiBaseUrl`. If you add:
  ```json
  {
    "apiBaseUrl": "https://gateway.example.com/v1",
    "promptSuffix": "\n(added from config)"
  }
  ```
  every API run (including stored-session replays) automatically uses that endpoint unless you override it with `--base-url`. CLI flag > config > `OPENAI_BASE_URL` env remains the precedence, and sessions record the chosen URL so `oracle session <id>` replays against the same gateway.
- Base URL handling now flows through `buildRunOptions`, `runOracle`, session metadata, and the client factory. Type safety improved by moving `ClientFactory` to `src/oracle/types.ts`, and all normalization lives in `src/cli/options.ts`.
- README + docs now explain the flag/config knobs and show the “Using OpenAI base URL …” log line so users know the override stuck. Tests cover config/env precedence and lint-safe env mocking.

Tested both CLI argument + config.json file on both GPT 5.1 and GPT-5 Pro. 

Example run:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,9 @@ npx -y @steipete/oracle -p "Review the TS data layer" --file "src/**/*.ts" --fil
 # Mixed glob + single file
 npx -y @steipete/oracle -p "Audit data layer" --file "src/**/*.ts" --file README.md
 
+# Alternate base URL (LiteLLM, Azure, self-hosted gateways)
+OPENAI_API_KEY=sk-... oracle --base-url https://litellm.example.com/v1 -p "Summarize the risk register"
+
 # Inspect past sessions
 oracle status --clear --hours 168   # prune a week of cached runs
 oracle status                       # list runs; grab an ID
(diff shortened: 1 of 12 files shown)


Tr1Fecta-7:
Not sure! I don't know how they handle authentication, but if it's a simple open ai url change with api key then it should. I tested with LiteLLM and it works as expected.

Tr1Fecta-7:
> Thanks, this looks useful! Will this work for Azure? I'll need an account to test.

According to this: https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/codex?view=foundry-classic&tabs=npm

You'd have to set the new base url argument to: `https://YOUR_RESOURCE_NAME.openai.azure.com/openai/v1`

And the OpenAI api key as your given azure api key then it works!

PR #30, opened 2025-12-06
fix(browser): use full mouse event sequence for send button

mikedemarais:
When using Oracle in browser mode, prompts would be correctly injected into ChatGPT's textarea—you could see the text appear—but then nothing would happen. Oracle would hang indefinitely, waiting for a response that never came because the send button was never actually triggered.

The root cause: ChatGPT's React-based UI doesn't respond to simple `.click()` calls. React's synthetic event system requires the full pointer/mouse event sequence (pointerdown → mousedown → pointerup → mouseup → click) to properly trigger event handlers. The codebase already uses this pattern for the "Answer now" button, but `attemptSendButton` was still using `.click()`.

Changes:
- Update `attemptSendButton` to dispatch the full mouse event sequence
- Add InputEvent dispatch after setting editor.textContent so React/ProseMirror recognizes the injected text

### DIFF
--- a/src/browser/actions/promptComposer.ts
+++ b/src/browser/actions/promptComposer.ts
@@ -106,6 +106,8 @@ export async function submitPrompt(
         const editor = document.querySelector(${primarySelectorLiteral});
         if (editor) {
           editor.textContent = ${encodedPrompt};
+          // Nudge ProseMirror to register the textContent write so its state/send-button updates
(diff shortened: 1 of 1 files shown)


steipete:
Thanks, love a focussed commit with a great message. I'll test this and do a new release soon.

Also might extract this into a helper, seems a good idea for any click.

PR #41, opened 2025-12-16
fix(browser): strip markdown code fences before prompt verification

mulyoved:
## Summary
- Fix prompt commit verification failing when prompts contain markdown code fences
- ChatGPT's UI renders/strips markdown differently than raw text, causing exact-match comparison to fail even when prompt was successfully sent
- Enhanced the `normalize` function to strip markdown code fences (` ``` `) and inline code (`` ` ``) before comparison

## Test plan
- [x] Tested with GPT-5.2 Pro via remote browser mode with file attachments
- [x] Verified prompts with markdown code blocks now succeed
- [x] All unit tests pass (470 passed, PTY tests skipped due to Windows native module issues)

### DIFF
--- a/src/browser/actions/attachments.ts
+++ b/src/browser/actions/attachments.ts
@@ -247,25 +247,43 @@ export async function waitForAttachmentCompletion(
     const filesAttached = attachedNames.length > 0;
     return { state: button ? (disabled ? 'disabled' : 'ready') : 'missing', uploading, filesAttached, attachedNames };
   })()`;
-  while (Date.now() < deadline) {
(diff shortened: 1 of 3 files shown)


steipete:
Thanks again for this, @mulyoved — merged.

We landed the overall improvements (markdown-safe prompt verification, stronger echo recovery incl. remote mode, and more reliable multi-file uploads) and added a couple safeguards to avoid false-positives (e.g. fence-marker stripping without deleting code content, and tighter echo/filename heuristics).

Also credited you in the changelog (0.7.1 Unreleased).

PR #54, opened 2025-12-30
Fix Thinking Time pill detection and make more cli flags configurable

TheCrazyLex:
The thinking time chip detection now also matches composer pills
labeled 'Pro' (in addition to 'Thinking'), fixing selection on UIs
where OpenAI uses the Pro branding for the thinking level menu.

Also some more support for config.json:
- browser.thinkingTime is now supported
- browser.manualLogin + browser.manualLoginProfileDir are now supported
- Additional Unit test coverage


steipete:
Thanks, taking over!

steipete:
Merged manually on main (fast-forwarded PR commits, then doc tweak in main).

PR #90, opened 2026-02-07
fix(browser): resolve TDZ crash in markdown fallback extractor

julianknutsen:
## Summary

- `buildMarkdownFallbackExtractor('MIN_TURN_INDEX')` generates `const MIN_TURN_INDEX = (MIN_TURN_INDEX >= 0 ? ...)` — a `const` that references itself in its own initializer (temporal dead zone)
- This crashes the DOM observer expression on **every request**, forcing all responses through `recoverAssistantResponse` which grabs the first text with no stability checks
- Renamed the inner variable to `__minTurn` to avoid shadowing the outer scope

## Impact

The bug affects all browser-mode requests where `minTurnIndex` is passed (which is every normal prompt submission). The DOM observer path silently crashes with `ReferenceError: Cannot access 'MIN_TURN_INDEX' before initialization`, and the recovery fallback returns truncated output.

**Before fix:** ~1.7k tokens captured (truncated mid-word), every run shows "Recovered assistant response via polling fallback"
**After fix:** ~5.8k tokens captured (complete response with proper ending)

## Test plan

- [x] Tested with GPT-5.2 Thinking (heavy) — short prompt (5 sentences): complete response
- [x] Tested with GPT-5.2 Thinking (heavy) — long prompt (5000-word essay): 23,670 chars, complete with conclusion
- [x] Verified no "Recovered assistant response via polling fallback" in verbose output
- [x] Verified DOM observer path now succeeds ("Refreshed assistant response via latest snapshot")
- [ ] Run existing test suite (`pnpm test`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- Browser: fix markdown fallback extractor TDZ crash in browser mode (#90, thanks @julianknutsen).
+
 ## 0.8.5 — 2026-01-19
 
 ### Added
--- a/tests/browser/pageActionsExpressions.test.ts
+++ b/tests/browser/pageActionsExpressions.test.ts
@@ -21,10 +21,18 @@ describe('browser automation expressions', () => {
 
   test('markdown fallback filters user turns and respects assistant indicators', () => {
     const expression = buildMarkdownFallbackExtractorForTest('2');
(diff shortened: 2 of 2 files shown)


steipete:
Landed via temp rebase onto main.

- Gate: pnpm lint && pnpm build && pnpm test && pnpm docs:list
- Land commit (PR head): 4b399091
- Merge commit: 89bbc41c

Thanks @julianknutsen!

PR #93, opened 2026-02-12
fix(browser): increase post-capture polling budget for thinking models

### LINKED ISSUE #85: Browser mode truncates long thinking model responses (post-capture verification too short)
## Problem

Browser mode truncates responses from GPT-5.2 Thinking and similar models when they pause mid-stream to "think" during markdown rendering.

The post-capture verification in `src/browser/index.ts` only waits ~5.5s max (1.5s initial + 5×800ms re-polls):

```typescript
await delay(1500);
let bestLength = capturedLength;
let bestText = answerText;
for (let i = 0; i < 5; i++) {
  // ...
  await delay(800);
}
```

For thinking models doing large code block outputs, this isn't enough. The model can pause for 10-30+ seconds mid-response while rendering, causing premature capture.

## Observed behavior

- GPT-5.2 Thinking response gets truncated mid-sentence/mid-codeblock
- Output file shows partial response
- Browser shows full response after oracle captures

## Expected behavior

Wait longer for thinking models, or add configurable timeout.

## Suggested fix

Option A: Add env var `ORACLE_POST_CAPTURE_TIMEOUT_MS` (default 5500, recommend 60000 for thinking models)

Option B: Detect thinking models and use longer timeout automatically (e.g., 60s for GPT-5.2 Thinking)

Option C: Add `--browser-post-capture-timeout` CLI flag

## Environment

- oracle 0.8.5
- macOS
- GPT-5.2 Thinking via browser mode
- Large context (~500k tokens) prompts

0xRaini:
Closes #85

Thinking models (GPT-5.2 Thinking, o-series) can pause mid-stream for 10-30s during extended reasoning. The post-capture verification only waited ~5.5s max (`1.5s + 5×800ms`), causing truncation.

**Changes:**
- Max iterations: 5 → 15 (~45s total budget)
- Require 3 consecutive stable polls before declaring done (handles temporary thinking pauses)
- Exponential backoff: 800ms → 4s during stable periods to reduce CPU waste
- Reset delay immediately when new content arrives

The original fast path (short responses ≤500 chars) is unchanged.

### DIFF
--- a/src/browser/index.ts
+++ b/src/browser/index.ts
@@ -1045,15 +1045,26 @@ async function maybeRecoverLongAssistantResponse({
   await delay(1500);
   let bestLength = capturedLength;
   let bestText = answerText;
-  for (let i = 0; i < 5; i++) {
+  // Thinking models (GPT-5.2 Thinking, o-series) can pause mid-stream for 10-30s.
+  // Use up to 15 iterations with increasing delays (total budget ~45s) to avoid truncation.
+  let pollDelay = 800;
+  let stableCount = 0;
+  for (let i = 0; i < 15; i++) {
     const laterSnapshot = await readAssistantSnapshot(runtime, baselineTurns ?? undefined).catch(() => null);
     const laterText = typeof laterSnapshot?.text === 'string' ? laterSnapshot.text.trim() : '';
     if (laterText.length > bestLength) {
       bestLength = laterText.length;
       bestText = laterText;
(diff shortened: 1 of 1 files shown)


0xRaini:
Hey @steipete -- the previous 5.5s polling budget was too short for thinking models that can pause 10-30s mid-stream. This extends the budget to ~45s with exponential backoff, requiring 3 consecutive stable reads before stopping.

PR #137, opened 2026-04-16
fix: remove overly broad endsWith guard that double-starts MCP server

SyntaxSmith:
## Summary
- When `oracle-mcp` is invoked via nvm/npm bin symlink, `process.argv[1]` resolves to a path like `.../bin/oracle-mcp` which matches the `endsWith("oracle-mcp")` guard in `server.ts:43`
- This triggers `startMcpServer()` twice — once from `bin/oracle-mcp.ts` and once from the module-level guard — creating two MCP server instances on the same stdin/stdout
- Every JSON-RPC request gets two identical responses, violating the protocol and causing MCP clients (Claude Code, etc.) to immediately disconnect
- Fix: remove the `endsWith("oracle-mcp")` fallback. The `import.meta.url` check alone covers the standalone-script case, and `bin/oracle-cli.ts` (lines 9–12) already handles `npx @steipete/oracle oracle-mcp`

## Test plan
- [x] `echo '<init>' | oracle-mcp` returns 1 response line (was 2 before fix)
- [x] `echo '<init>' | oracle oracle-mcp` still returns 1 response line (unaffected)
- [x] `pnpm test:mcp:unit` — 5 files, 7 tests passed


steipete:
Superseded by #160, which landed the focused oracle-mcp double-start fix with regression coverage and green macOS/Ubuntu/Windows CI, without the unrelated remote queue / MCP policy changes. Thanks @SyntaxSmith.

PR #146, opened 2026-04-30
Fix ChatGPT composer-pill DOM rewrite (model + thinking-effort)

SyntaxSmith:
## Summary

Restores browser mode after ChatGPT's recent composer rewrite:

- The model picker dropped \`data-testid=\"model-switcher-dropdown-button\"\` in favor of a \`__composer-pill\` button. \`MODEL_BUTTON_SELECTOR\` now matches both the legacy testid and the new pill so existing accounts and the rewritten composer both work.
- The standalone "Thinking" composer chip is gone; effort is now a per-row trailing button inside the model menu (\`[data-model-picker-thinking-effort-action=\"true\"]\`). \`buildThinkingTimeExpression\` opens the model menu, picks the trailing button for the selected row, resolves the submenu via \`aria-controls\`, and selects the matching effort, with bilingual EN/中文 label matching for users on Chinese ChatGPT. The old composer-chip flow stays as a fallback.
- \`ensureThinkingTime\` now logs a debug DOM dump and returns on any not-found state instead of throwing. With ChatGPT's UI churning, the safer default is to keep Pro consults alive on whatever effort the UI defaults to rather than fail the run.

Without this, every browser-mode consult fails at \"Unable to locate the ChatGPT model selector button\" (or, after working around that, at \"Unable to find the Thinking chip button in the composer area\").

## Test plan

- [x] \`pnpm build\`
- [x] \`pnpm lint\` (typecheck + oxlint, 0 errors)
- [x] \`pnpm test\` (620 passed / 41 skipped)
- [x] Live ChatGPT browser smoke against \`gpt-5.4-pro\` via \`oracle serve\` + MCP \`consult\`: returned answer in 44.9s, no errors thrown by model picker or thinking-time selection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Browser: track ChatGPT's April 2026 composer rewrite — the model picker dropped its `data-testid="model-switcher-dropdown-button"` in favor of a `__composer-pill` button, and the per-effort "Thinking" chip moved into a per-row trailing button inside the model menu (`[data-model-picker-thinking-effort-action]`). Model selection now matches both the legacy testid and the new pill, and thinking-time selection navigates the new model-menu trailing button (with bilingual EN/中文 label matching) before falling back to the old composer chip. When neither layout is available, thinking-time selection now degrades to a logged warning instead of failing the run, so Pro consults still work while the UI keeps shifting.
+
 ## 0.9.0 — 2026-03-08
 
 ### Changed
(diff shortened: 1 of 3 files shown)


needsmoretime:
Thank you for this. Do you know how to have it select Configure so that older models like 5.4 Pro can be used? I wasn't able to get that working.

SyntaxSmith:
> Thank you for this. Do you know how to have it select Configure so that older models like 5.4 Pro can be used? I wasn't able to get that working.

No,I have no idea. Maybe you can choose it manually and this tool will use it by default

steipete:
Superseded by #158, which landed the focused ChatGPT composer-pill model selector and thinking-effort menu fix from this PR, with updated tests and CHANGELOG credit. Thanks @SyntaxSmith.

PR #154, opened 2026-05-01
fix(gemini-web): prefer latest non-empty stream chunk in parser

### LINKED ISSUE #153: Bug: Gemini browser mode returns (no text output) for gemini-3-pro / gemini-3.1-pro
## Summary

When running `oracle --engine browser --model gemini-3-pro` (or `gemini-3.1-pro`), the CLI consistently prints `Answer: (no text output)` with `↓0` tokens, even though the request actually succeeds and a conversation is created at `gemini.google.com/app/<cid>` containing a complete answer.

The bug appears to live in oracle's response parser for the Gemini Web `StreamGenerate` endpoint. The server streams the response across multiple `wrb.fr` chunks; the **first** chunk that carries a candidate is usually an empty placeholder (`text = ""`), and the actual answer arrives in subsequent chunks. The parser stops at that placeholder and discards the rest, so the user-facing CLI shows no text.

The bug reproduces consistently for `gemini-3-pro` / `gemini-3.1-pro` on Windows 11 through the browser engine. I have anecdotal reports that the same command does not always surface the bug for macOS users; the underlying reason is unconfirmed (the parser is sensitive to the number of `wrb.fr` chunks in the streamed response — see "Why" below — but I have not measured whether macOS systematically receives fewer chunks). Whatever the cause, environments that do see multi-chunk responses hit the bug; environments that see a single coalesced chunk do not.

## Environment

- oracle 0.9.0
- Node 24+ (per the project's current toolchain)
- Windows 11; not reproduced on macOS (see "Why" below)
- Chrome profile already signed in to `gemini.google.com`, reused via `--browser-manual-login`

## Repro

```bash
oracle --engine browser --browser-manual-login \
  --model gemini-3-pro -p "ping" --file README.md --force
```

> PowerShell: replace `\` with backtick `` ` `` or join the command into a single line.

### Observed

```
🧿 oracle 0.9.0 — ...
Using Gemini web client for browser automation
Launching browser mode (gemini-3-pro) with ~4,904 tokens.
Answer:
(no text output)

17.2s · gemini-3-pro[browser] · ↑4.9k ↓0 ↻0 Δ4.9k
files=1
```

### Expected

A non-empty answer rendered after `Answer:`, matching what is visible at `gemini.google.com/app/<cid>` for the same conversation.

### Cross-check

Opening `gemini.google.com/app/<cid>` (the tab oracle just created) shows that Gemini did return a full response (e.g. `"Pong! 🏓 …"`). The conversation exists and the server has committed it; only the client side fails to extract the text.

## Why

The Gemini Web `StreamGenerate` endpoint returns a concatenated JSON payload (prefixed with the standard `)]}'` XSSI-prevention guard) made of multiple `["wrb.fr", null, "<JSON string>"]` blocks flushed progressively. For `gemini-3-pro` / `gemini-3.1-pro`, the server emits ~7 chunks within a single response body. The value of `body[4][0][1][0]` (the text candidate) across the chunks looks like:

```text
Chunk 1: (no [4] — control chunk "Acknowledge Initial Context")
Chunk 2: ""                                             (placeholder, rcid present)
Chunk 3: "Pong! 🏓\n\nI've received the `README"
Chunk 4: "...interacting with various LLMs (like"
Chunk 5: "...troubleshooting an issue, drafting"
Chunk 6: "...or setting up an integration?"             (full text)
Chunk 7: full text + finalize flag
```

The current parser stops at Chunk 2, takes `text = ""`, and falls back to `(no text output)`.

When the response is coalesced into a single chunk (whatever the cause — could be route, region, or model-specific server behavior), the first chunk already carries the full text and the bug never surfaces. This is why the bug is intermittent and likely why CI has not caught it.

### Captured raw response

A full 48 KB raw response payload is available locally and can be shared on request — it confirms the chunk sequence above (each `wrb.fr` block parsed and the text growing monotonically chunk by chunk).

## Code references (verified on `main`)

These pointers are provided to help anyone investigating; they are not a prescription for how to fix it.

- `src/gemini-web/client.ts:251-265` — the `for` loop inside `parseGeminiStreamGenerateResponse` `break`s at the first chunk where `candidateList.length > 0`, with no consideration for empty text. This is where the text is selected.
- `src/gemini-web/client.ts:271` — `textRaw = getNestedValue<string>(firstCandidate, [1, 0], "")`. Once the parser locks onto a `[""]` chunk, the function returns `text: ""`.
- `src/gemini-web/client.ts:384-412` — `runGeminiWebOnce` wraps the parse in a try/catch. Because the parser does not throw (it just returns empty text), the call returns `text: ""` through the happy path with no error signal — this is why the failure is silent.
- `src/gemini-web/executor.ts:536-556` — `answerText = response.text ?? ""` propagates downstream; the output token count is computed from the empty `answerText`, which is why the log shows `↓0`.
- `src/browser/sessionRunner.ts:125-128` — the `(no text output)` fallback fires when both `answerMarkdown` and `answerText` are falsy.

## Notes

- I will follow up in a comment with a candidate fix and its tradeoffs, kept separate from this report so maintainers or other contributors can suggest a different direction first.
- Happy to run additional repros, share the captured raw payload, or test a different fix direction if preferred.

manhtruong03:
# Summary

Fixes the `(no text output)` symptom for `gemini-3-pro` / `gemini-3.1-pro` browser-engine runs by making `parseGeminiStreamGenerateResponse` walk every `wrb.fr` chunk and prefer the latest one with non-empty text, instead of `break`-ing at the first chunk with a candidate (which is usually an empty placeholder under streaming).

Fixes #153.

## What changed

- `src/gemini-web/client.ts` — two adjustments inside `parseGeminiStreamGenerateResponse`:
  - drop the early `break` in the chunk-selection loop; iterate to the end and pick the latest chunk whose text is non-empty. The first chunk with a candidate is kept as the anchor so coalesced 1-chunk responses still parse.
  - switch the candidate-text read from `getNestedValue<string>(...)` to `getNestedValue<unknown>(...)` and narrow with `typeof === "string"`. The helper does not validate the runtime shape; this matches the pattern already used elsewhere in the file (lines 256, 269) and prevents a silent lock-in if the schema drifts.
- `tests/gemini-web/parse.test.ts` — 3 new unit tests covering multi-chunk streaming, single-chunk regression guard, and a defensive empty-after-non-empty case. Two small synthetic-body helpers added to keep the new cases readable.

Diff size: +56 / -7 across 2 files.

## Why this approach

- **Smallest possible change** that resolves the symptom. Buffer-and-concat would risk double-printing because each chunk is already cumulative on the wire.
- **Preserves existing single-chunk behavior** (the route that masks this bug on macOS) — that path was working and stays unchanged.
- **Image-generation branch — not exercised in this PR.** The image extraction loop (`src/gemini-web/client.ts:297-328`) iterates from `bodyIndex` forward looking for chunks with image data. With the new predicate, `bodyIndex` is the latest chunk with non-empty text instead of the first chunk with a candidate. In every captured response the image-bearing chunks arrive at-or-after the final text chunk, so the loop still finds them — but I do not have an automated test for this and a maintainer with `--generate-image` access may want to spot-check before merging.

## How to test

```bash
pnpm install
pnpm typecheck
pnpm vitest run tests/gemini-web        # 18 / 18 pass (15 existing + 3 new)
pnpm vitest run                         # 615 pass / 0 fail / 49 skip

# End-to-end (needs a Gemini account already signed in via --browser-manual-login)
pnpm start --engine browser --browser-manual-login \
  --model gemini-3-pro -p "ping" --file README.md --force
```

Expected end-to-end: a non-empty `Answer:` block and `↓<n>` tokens > 0 (was `↓0` before the fix).

## Test plan

- [x] `pnpm typecheck` passes.
- [x] `pnpm vitest run tests/gemini-web` passes (18/18).
- [x] Full suite `pnpm vitest run` passes (615/0/49).
- [x] Manual end-to-end repro on Windows 11 / Node 25, browser engine, `gemini-3-pro` — answer text renders, output tokens > 0 (was `↓0` before the fix).
- [ ] Cross-platform validation deferred to CI; I do not have macOS/Linux to run locally. If CI does not exercise the browser engine path, I am happy to add one or to debug failures the maintainer surfaces.
- [ ] `--generate-image` branch not exercised — see "Risks" below.

## Risks and non-goals

- **No public API change.** Parser shape is internal; only `text` extraction logic moves.
- **`bodyIndex` semantic shift.** Before: first chunk with a candidate. After: the chunk eventually selected as `body` (typically the last with non-empty text). Only one in-function consumer (`hasGenerated`) reads `bodyIndex`; behavior verified for the text path, not actively tested for the image path. Documented as a known limitation in the issue comment.
- **Trailing-error-chunk edge case.** The new predicate prefers the last chunk with non-empty text. If the server ever ships a final chunk whose text is a server-side error string, it would silently override the real answer. Not observed in captured payloads; will harden if it surfaces.
- **Out of scope:** adding a `log?.warn(...)` when `res.ok` but `text === ""` — separate follow-up PR. "Longest text" predicate also out of scope (no data justifying it yet).
- **No new dependencies, no config changes, no migrations.**

## Notes for the reviewer

- The `typeof` guard reads slightly defensive for a 4-line patch; rationale is in the inline comment and in the issue thread (the helper does not validate runtime types).
- A 48 KB raw response payload from a real `gemini-3-pro` run was used to derive the chunk sequence. It is not committed because (a) the synthetic minimal bodies in the new tests cover the same shapes more readably, and (b) the raw response embeds an account-bound conversation URL that I would need to scrub before sharing. I can attach a scrubbed copy to the issue if a reviewer wants it.

Branch: `fix/gemini-streaming-parse-empty-text` (single commit: `fix(gemini-web): prefer latest non-empty stream chunk in parser`).

### DIFF
--- a/src/gemini-web/client.ts
+++ b/src/gemini-web/client.ts
@@ -255,9 +255,15 @@ export function parseGeminiStreamGenerateResponse(rawText: string): {
       const parsed = JSON.parse(partBody) as unknown;
       const candidateList = getNestedValue<unknown[]>(parsed, [4], []);
       if (Array.isArray(candidateList) && candidateList.length > 0) {
-        bodyIndex = i;
-        body = parsed;
-        break;
+        // Streaming: later chunks extend earlier ones; the first chunk with a
+        // candidate is often an empty placeholder. Prefer the latest chunk
+        // whose text is non-empty.
+        const candidateText = getNestedValue<unknown>(candidateList[0], [1, 0], "");
+        const hasText = typeof candidateText === "string" && candidateText.length > 0;
+        if (body === null || hasText) {
+          bodyIndex = i;
+          body = parsed;
+        }
       }
     } catch {
       // ignore
--- a/tests/gemini-web/parse.test.ts
+++ b/tests/gemini-web/parse.test.ts
@@ -5,8 +5,23 @@ import {
 } from "../../src/gemini-web/client.js";
 
 function makeRawResponseWithBody(body: unknown): string {
-  const responseJson = [[null, null, JSON.stringify(body)]];
+  return makeRawResponseWithBodies([body]);
+}
+
+function makeRawResponseWithBodies(bodies: unknown[]): string {
+  const responseJson = bodies.map((body) => [null, null, JSON.stringify(body)]);
   return `)]}'\n\n${JSON.stringify(responseJson)}`;
+}
+
+function makeBodyWithText(rcid: string, text: string): unknown[] {
+  const candidate: unknown[] = [];
+  candidate[0] = rcid;
+  candidate[1] = [text];
+
+  const body: unknown[] = [];
+  body[1] = ["cid", "rid", rcid];
+  body[4] = [candidate];
+  return body;
 }
 
 describe("gemini-web parseGeminiStreamGenerateResponse", () => {
@@ -66,6 +81,34 @@ describe("gemini-web parseGeminiStreamGenerateResponse", () => {
 
     const parsed = parseGeminiStreamGenerateResponse(makeRawResponseWithBody(body));
     expect(parsed.text).toBe("Expanded card content");
+  });
+
+  it("picks the latest non-empty text across streaming chunks", () => {
+    const raw = makeRawResponseWithBodies([
+      makeBodyWithText("rcid-1", ""),
+      makeBodyWithText("rcid-1", "Pong! 🏓 partial"),
+      makeBodyWithText("rcid-1", "Pong! 🏓 partial answer with more"),
+      makeBodyWithText("rcid-1", "Pong! 🏓 partial answer with more — final."),
+    ]);
+
+    const parsed = parseGeminiStreamGenerateResponse(raw);
+    expect(parsed.text).toBe("Pong! 🏓 partial answer with more — final.");
+  });
+
+  it("still parses a single coalesced chunk (regression guard)", () => {
+    const raw = makeRawResponseWithBody(makeBodyWithText("rcid-1", "single-chunk reply"));
+    const parsed = parseGeminiStreamGenerateResponse(raw);
+    expect(parsed.text).toBe("single-chunk reply");
+  });
+
+  it("preserves the previous non-empty chunk when a later chunk has empty text", () => {
+    const raw = makeRawResponseWithBodies([
+      makeBodyWithText("rcid-1", "first answer"),
+      makeBodyWithText("rcid-1", ""),
+    ]);
+
+    const parsed = parseGeminiStreamGenerateResponse(raw);
+    expect(parsed.text).toBe("first answer");
   });
 
   it("extracts model-unavailable error code 1052 from response json", () => {


steipete:
Superseded by #157, which landed the same Gemini streaming parser fix with regression tests and kept the generated-image scan anchored at the first candidate chunk. Thanks again @manhtruong03; credited in CHANGELOG.md.

PR #173, opened 2026-05-05
feat(browser): manage ChatGPT Project Sources

pdurlej:
## Summary

Adds a focused, non-destructive ChatGPT Project Sources v1 for browser-mode workflows:

- `oracle project-sources list|add` for the ChatGPT Project Sources tab.
- MCP `project_sources` tool with `operation:"list"|"add"`.
- `add` supports `--dry-run` / `dryRun:true` and MCP requires `confirmMutation:true` for persistent mutations.
- Project Sources browser automation does not select a model, start a consult, or send a prompt.

This addresses #131 and is a narrower successor/alternative to #132. Credit to @vgorlovi for identifying the Project Sources workflow and proving the first spike. This PR keeps only the safe v1 surface: list, append, and dry-run. Delete/replace/sync are intentionally left for a follow-up after more UI coverage.

## Why

With ChatGPT Developer Mode / Memory Off, separate chats in a Project do not implicitly share chat history. Project Sources still provide explicit shared project context through uploaded files. That makes Sources useful for agent workflows where context should be deliberate and inspectable rather than hidden memory.

## Safety / scope

- Browser only; no API model ids or model picker changes.
- Local signed-in browser host only for v1; remote `oracle serve` support is rejected for `project_sources` for now.
- No destructive source operations.
- Mutating MCP calls require `confirmMutation:true` unless `dryRun:true`.
- Upload input selection is scoped to the recognized Add Sources dialog to avoid normal composer upload controls.

## Tests

Local:

```text
pnpm run check
# All matched files use the correct format.
# tsgo --noEmit
# Found 0 warnings and 0 errors.

pnpm run build
# tsgo -p tsconfig.build.json && pnpm run build:vendor

pnpm vitest run tests/projectSources/url.test.ts tests/browser/projectSources.test.ts tests/cli/projectSources.test.ts tests/mcp/projectSources.test.ts
# Test Files 4 passed (4)
# Tests 13 passed (13)

pnpm test
# Test Files 118 passed | 18 skipped (136)
# Tests 790 passed | 41 skipped (831)

git diff --check
# no output
```

CLI dry-run:

```text
node ./dist/bin/oracle-cli.js project-sources add \
  --chatgpt-url https://chatgpt.com/g/g-p-123/project \
  --file /tmp/oracle-project-sources-dry-run.md \
  --dry-run --json

status: dry-run
projectUrl: https://chatgpt.com/g/g-p-123/project?tab=sources
plannedUploads: 1
```

[...]

PR #179, opened 2026-05-05
feat(browser): explain browser control before runs

pdurlej:
## Summary
- add a small browser control-plan helper so dry-runs and live browser sessions say whether Oracle will launch visible Chrome, hide a new window, attach to an existing browser, or use remote/headless Chrome
- surface those browser-control lines through non-verbose session logs so MCP/agent callers get a visible progress signal before long GPT-5.5 Pro waits
- document calmer shared-desktop/Claude Code usage and clarify that long browser consults should be inspected via sessions/status before retrying
- let `--browser-attach-running` ignore conflicting launch-only defaults from user config, while still allowing explicitly provided conflicts to fail validation

## Why
Agents such as Claude Code can look stuck during long browser-backed GPT-5.5 Pro consults, and visible Chrome automation can surprise the operator. This PR keeps behavior explicit before touching ChatGPT and makes the attach-running path usable even when a user config defaults to manual-login browser launches.

## Tests
- `pnpm install --frozen-lockfile`
- `pnpm vitest run tests/cli/browserDefaults.test.ts tests/cli/browserConfig.test.ts tests/cli/dryRun.test.ts tests/cli/dryRun.coverage.test.ts tests/browser/controlPlan.test.ts tests/browser/sessionRunner.test.ts`
- `pnpm run check`
- `pnpm run build`
- `pnpm test`
- `git diff --check`

## Smoke / Dogfood
- `node ./dist/bin/oracle-cli.js --dry-run summary --engine browser --model gpt-5.5-pro --browser-attach-running -p "CHECK_CONTROL_PLAN" --file README.md`
  - confirmed preview prints `Browser control: attach to an already-running local Chrome session` without inheriting manual-login config conflicts
- Oracle GPT-5.5 Pro browser review, slug `calmer-browser-ux-review`: `ACCEPTABLE`
- Oracle GPT-5.5 Pro browser follow-up, slug `calmer-browser-ux-review-2`: `ACCEPTABLE`; during the wait it emitted `[browser] Waiting for ChatGPT response ...`, matching the agent-facing progress case this PR documents

PR #178, opened 2026-05-05
feat(browser): auto-archive one-shot ChatGPT runs

pdurlej:
## Summary

- Adds `browser.archiveConversations` / `--browser-archive <auto|always|never>` / MCP `browserArchive` for ChatGPT browser runs.
- Default `auto` archives only successful non-project, non-Deep-Research, non-multi-turn one-shot chats after local artifacts are saved.
- Keeps archive failures non-fatal, but avoids false success: archive only runs after required artifacts are saved and reports `archived: true` only after UI confirmation / URL change.
- Stores archive result metadata with the browser session.

## Safety / Scope

- Failed, incomplete, project, Deep Research, and multi-turn runs are not auto-archived in `auto`.
- `always` can force successful browser conversations, including Deep Research, but still waits for report/transcript persistence first.
- The DOM action is scoped to visible ChatGPT menu items so conversation titles containing “archive” are not clicked by mistake.
- Archived chats remain manageable in ChatGPT; local Oracle transcript/artifacts remain the durable record.

## Tests

- `pnpm vitest run tests/browser/archiveConversation.test.ts tests/browser/index.test.ts tests/browser/config.test.ts tests/cli/browserConfig.test.ts tests/cli/dryRun.test.ts tests/browser/sessionRunner.test.ts tests/mcp/consult.test.ts`
- `pnpm run check`
- `pnpm run build`
- `pnpm test` — 114 files passed, 18 skipped; 785 tests passed, 41 skipped
- `git diff --check`

## Dogfood / Live Smoke

- Oracle GPT-5.5 Pro review #1 returned `NOT ACCEPTABLE` and caught three blockers: archive after artifact save failure, unverified archive success, and missing Deep Research `always` wiring. All were fixed.
- Oracle GPT-5.5 Pro follow-up review returned `ACCEPTABLE`.
- Live ChatGPT browser smoke:
  - `archive-smoke-final-ok`
  - command shape: browser mode, `gpt-5.5-pro`, `--browser-thinking-time extended`, `--browser-archive always`, one README attachment
  - final answer contained `CHECK_ARCHIVE_FINAL_OK`
  - local render succeeded with transcript artifact at `/Users/pd/.oracle/sessions/archive-smoke-final-ok/artifacts/transcript.md`
  - session metadata recorded `archive: { mode: "always", attempted: true, archived: true }`

No secrets, cookies, or account-specific payloads are included in this PR.

PR #181, opened 2026-05-05
feat(browser): recommend conversation mode in dry-runs

pdurlej:
## Summary
- add an advisory browser conversation-mode recommendation for dry-runs: `one-shot`, `multi-turn`, `deep-research`, or `project`
- surface the recommendation in CLI dry-run output, browser preview JSON, and MCP `consult` dry-run structured content as optional `recommendedConversationMode`
- document that recommendations are guidance only and never switch projects, invent follow-ups, upload Project Sources, or change archive behavior

## Safety / Scope
- no browser DOM, model picker, composer, or Project Sources implementation changes
- no dependency on #173; Project Sources are mentioned only as persistent ChatGPT Project context in docs
- existing dry-run/preview fields keep their meaning; MCP field is optional

## Tests
- `pnpm install --frozen-lockfile`
- `pnpm vitest run tests/cli/conversationMode.test.ts tests/cli/dryRun.test.ts tests/cli/dryRun.coverage.test.ts tests/mcp/consult.test.ts` -> 4 files passed, 24 tests passed
- `pnpm run check`
- `pnpm run build`
- `pnpm test` -> 116 files passed, 18 skipped; 796 tests passed, 41 skipped
- `git diff --check`

## Notes
This is intended to follow the multi-turn guardrails work, but it remains independently reviewable. If #180 lands first, I can rebase this PR to pick up the wording and resolve any small docs/dry-run conflicts.


pdurlej:
Rebased onto current `main` after #178, #179, and #180 landed. Conflict resolution preserved the new browser-control/multi-turn guardrail output and added the advisory conversation-mode recommendation after the control plan, without changing runtime behavior.

Local verification after rebase:

```text
pnpm vitest run tests/cli/conversationMode.test.ts tests/cli/dryRun.test.ts tests/cli/dryRun.coverage.test.ts tests/mcp/consult.test.ts
# 4 passed, 27 tests passed

pnpm run check
pnpm run build
pnpm test
# 118 passed, 18 skipped; 823 tests passed, 41 skipped

git diff --check
```

GitHub CI is green on Ubuntu, macOS, and Windows.

steipete:
Thanks for rebasing this and for keeping the behavior advisory-only. I took another pass now that #173 landed. I am going to close this one rather than continue the stack.

Reasoning:

- The useful guardrail from this stack already landed in #180: browser follow-ups are explicit only, and Oracle does not invent multi-turn prompts.
- #173 now gives Project Sources a real command surface (`project-sources list|add`) instead of a dry-run recommendation.
- This PR adds a prompt-content heuristic (`project` for prompts that look like ongoing architecture/product streams). That is plausible, but it is also easy for a CLI to become noisy or over-directive when it guesses workflow mode from prose.
- For agent workflows, I would rather keep dry-runs factual: what Oracle will do, which browser/project URL it will use, which follow-ups/files/sources were explicitly provided, and what will or will not mutate state.

The design may still be useful later if we want an explicit `oracle plan`/advisor mode or if MCP clients ask for workflow recommendations separately from dry-run execution. I do not think it belongs in the core dry-run path right now.

Appreciate the careful scope: no browser DOM changes, optional MCP field, and clear docs. This is being closed as product-direction deferral, not because the implementation was careless.

PR #220, opened 2026-05-19
fix: keep polling send button for attachments

joanvelja:
## What changed

- Keep probing ChatGPT's send button during attachment submissions even when the attachment-chip readiness detector has not fired yet.
- Continue waiting for a clickable send button for attachment submissions instead of falling back to Enter.
- Add a regression test for a ready send button with a still-pending attachment readiness signal.

## Why

ChatGPT can enable the send button after attachments finish while Oracle's DOM-based chip detector still reports pending. The old flow skipped the send-button probe until that detector returned true, so browser submissions could hang instead of clicking send.

## Validation

- `pnpm vitest run tests/browser/promptComposer.test.ts tests/browser/promptComposerExpressions.test.ts`
- `pnpm run check`

### DIFF
--- a/src/browser/actions/promptComposer.ts
+++ b/src/browser/actions/promptComposer.ts
@@ -483,22 +483,24 @@ async function attemptSendButton(
   // settle slowly for multi-file uploads, but plain text sends should keep the
   // shorter historical deadline.
   const deadline = Date.now() + sendButtonTimeoutMs(attachmentNames);
+  let attachmentReady = !needAttachment;
   while (Date.now() < deadline) {
     if (needAttachment) {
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for the PR. I landed the attachment readiness fixes through #222 and #212 instead.

Those commits anchor readiness to the active ChatGPT composer, handle renamed and ellipsized attachment chips, and keep the checks scoped to attachment affordances. Proof: focused browser attachment tests, pnpm run check, pnpm run build, clean autoreview, and green GitHub checks on #212.

This PR now overlaps the landed path and conflicts with current main, so I am closing it rather than carrying a second implementation.

PR #226, opened 2026-05-26
feat(session): recover --harvest/--live by reopening saved tab URL

hbruceweaver:
## Summary

- When `oracle session <id> --harvest` or `--live` fails to find a live ChatGPT tab, automatically relaunch Chrome with the session's manual-login profile, navigate to the conversation URL stored in session metadata, and retry the harvest against the recovered tab.
- Default on; opt out with `--no-recover` on the `session` subcommand.
- New module `src/browser/recoverConversation.ts`. Recovery URL is gated by `isRecoverableChatGptConversationUrl` (extracted from `hasRecoverableChatGptConversation`) so home, project shell, and external URLs in stale metadata can never navigate the persistent profile to the wrong page. Preference order: `meta.browser.harvest.url` then `meta.browser.runtime.tabUrl` (matches `resolveSessionTabRef`).
- Profile dir prefers `meta.browser.config.manualLoginProfileDir`, falling back to `defaultManualLoginProfileDir()`.
- The recovered Chrome is owned by a single try/finally that wraps both the recovery retry and (for `--live`) the main poll loop, so `closeAfterRecover` is honored even if the post-recovery harvest itself throws.
- Tests in `tests/browser/recoverConversation.test.ts` (10 cases on URL validation/preference) and `tests/cli/browserTabsRecover.test.ts` (3 cases on fallback path, opt-out, cleanup).

## Why

Long GPT-5.x Pro Extended runs frequently complete in the background after the CLI's 20-minute wall expires. The original session has the full conversation URL persisted in `~/.oracle/sessions/<id>/meta.json` (`browser.runtime.tabUrl`), and ChatGPT preserves attachments and the assistant response at that URL, but `--harvest` and `--live` only look at *live* tabs in a running Chrome — when the original CLI exits, the browser closes and the response becomes unrecoverable through `oracle`.

The motivating concrete case: a 191k-token GPT-5.5 Pro Extended code review timed out at the 20-minute wall. The original CLI saved `meta.browser.runtime.tabUrl` but exited with only `"I"` captured. The actual 17.5KB review had completed in ChatGPT's background. With this patch, `oracle session <slug> --live --write-output ...` relaunches the profile, reopens the URL, harvests the now-complete response, and writes it to disk in ~10 seconds.

## Real behavior proof

End-to-end recovery against the timed-out session described above. Paths, conversation IDs, and the dynamic CDP port redacted; all other output is verbatim from the patched build.

```
$ node ./dist/bin/oracle-cli.js session <session-slug> --harvest --write-output /tmp/recovered.md

🧿 oracle 0.13.0 — From shruggy agents to shippable PRs.
No live ChatGPT tab matched session "<session-slug>". Attempting recovery by reopening the saved conversation URL.
[browser] Recovery: relaunching Chrome with profile /Users/redacted/.oracle/browser-profile and navigating to https://chatgpt.com/c/<conversation-id>
[browser] Recovery: Chrome listening on 127.0.0.1:<port>; tab loaded.
Session: <session-slug>
Target: C11E88428F3D01E0CDD12E13312AB7D6
State: completed
Model: (unknown)
URL: https://chatgpt.com/c/<conversation-id>
Assistant turns: 4
Signals: stop=no send=no
Last user: hermor-tab-codebase(1).zip ZipArchive ...
---
1. **End-state alignment — usable substrate, not end-state…**

   The code has the core pieces: `hermor_tab/mcp_server.py` exposes lifecycle tools and auto-registers `SUPPORTED_TOOL_NAMES` as MCP tools (`mcp_server.py:257–309`); `api.py` has the external enqueue endpoint (`api.py:151–165`); the extension executor polls and executes server commands…
```

`--live` (full streaming variant) recovered the complete 17.5KB markdown of the original review (all 5 requested sections + 7 follow-up recommendations) in ~10 seconds total — the same conversation the original 20-min CLI run had given up on with only `"I"` captured. The harvested output was written to disk via `--write-output`; the live tail terminated cleanly when ChatGPT's `stopExists: false` was detected on the first poll (= response was already complete in background).

## Verification

- `pnpm run check` — clean (typecheck + format + lint)
- `pnpm docs:check` — ok (85 flags, 6 files)
- `pnpm vitest run` — **1076/1076 pass** (10 new tests added in the second commit)
- Manual end-to-end against the timed-out session described above: see proof block.

## Changes between the first push and now (review pass)

The second commit `fix(session): gate recovery URL + tighten cleanup` addresses ClawSweeper's findings:

1. **Recovery URL is now validated.** Extracted `isRecoverableChatGptConversationUrl(candidate)` from `hasRecoverableChatGptConversation` in [`src/browser/reattachability.ts`](https://github.com/steipete/oracle/blob/feat/harvest-recover-via-saved-url/src/browser/reattachability.ts) and reuse it in `resolveRecoveryUrl`. Stale home URLs (`https://chatgpt.com/`), project shells (`/g/g-…`), and unrelated external URLs are now rejected — recovery falls through to the original missing-tab error rather than navigating the persistent profile to the wrong page.
2. **Preference order flipped.** `harvest.url` is checked before `runtime.tabUrl`, matching `resolveSessionTabRef`. This way a stale `runtime.tabUrl = https://chatgpt.com/` (shell) can no longer beat a valid `harvest.url = https://chatgpt.com/c/<id>` from a prior harvest.
3. **Cleanup gap closed.** Both `harvestSessionBrowserOutput` and `liveTailSessionBrowserOutput` now wrap the recovery retry (and, for `--live`, the main poll loop) inside the same try/finally as the `recoveredChrome` declaration. `closeAfterRecover: true` is now honored even if the post-recovery harvest itself throws.
4. **New test file** `tests/browser/recoverConversation.test.ts`: 10 cases covering accepted URLs, rejected shell/project/external URLs, malformed input, the harvest-vs-runtime preference order, and empty browser metadata.

## Notes

- The recovery launch still uses a minimal flag set (anti-detection + lang + no-first-run + use-mock-keychain). Kept narrow on purpose to avoid drifting from `launchChrome`'s production flag set in `chromeLifecycle.ts`. Happy to refactor to share a helper if you'd prefer one source of truth.
- No CLI flag was added for `closeAfterRecover` (kept it as an internal option); could expose `--close-after-recover` if useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Browser: `oracle session <id> --harvest` and `--live` now auto-recover when the original Chrome has been closed by relaunching the manual-login profile and reopening the saved conversation URL, then retrying the harvest against the recovered tab. Resolves the failure mode where a long GPT-5 Pro Extended response completed in the background after the CLI's 20-minute wall expired and the conversation was archived. Opt out with `--no-recover` on the `session` subcommand.
+- Browser: `oracle session <id> --harvest` and `--live` now auto-recover when the original Chrome has been closed by relaunching the manual-login profile and reopening the saved conversation URL, then retrying the harvest against the recovered tab. Resolves the failure mode where a long GPT-5 Pro Extended response completed in the background after the CLI's 20-minute wall expired and the conversation was archived. Recovery URL selection prefers `browser.harvest.url` over `browser.runtime.tabUrl` and is gated by a shared ChatGPT-conversation-URL check (rejects home, project shell, and external URLs so the persistent profile can't be navigated to the wrong page from stale metadata). Opt out with `--no-recover` on the `session` subcommand.
 
 ## 0.13.0 — 2026-05-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.13.1 — Unreleased
 
+### Added
+
+- Browser: `oracle session <id> --harvest` and `--live` now auto-recover when the original Chrome has been closed by relaunching the manual-login profile and reopening the saved conversation URL, then retrying the harvest against the recovered tab. Resolves the failure mode where a long GPT-5 Pro Extended response completed in the background after the CLI's 20-minute wall expired and the conversation was archived. Opt out with `--no-recover` on the `session` subcommand.
+
 ## 0.13.0 — 2026-05-22
 
 ### Added
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -1182,6 +1182,10 @@ program
     "--browser-tab <ref>",
     "Override the browser tab ref used for harvesting/live tail (current, target id, URL, or title substring).",
   )
+  .option(
+    "--no-recover",
+    "Do not relaunch Chrome to reopen the saved conversation URL when --harvest/--live finds no live tab.",
+  )
   .addOption(new Option("--clean", "Deprecated alias for --clear.").default(false).hideHelp())
   .action(async (sessionId, _options: StatusOptions, cmd: Command) => {
     const { handleSessionCommand } = await import("../src/cli/sessionCommand.js");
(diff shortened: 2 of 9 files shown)


hbruceweaver:
@clawsweeper re-review

PR #229, opened 2026-05-29
[codex] Make ChatGPT image generation easy for MCP agents

umutkeltek:
## Summary

- Add an agent-focused `chatgpt_image` MCP tool backed by Oracle's ChatGPT browser image path.
- Add image output support to `consult`, returning typed local `artifacts` and `images` metadata.
- Preserve contributor implementation and credit while merging current `main` and hardening the public contract.

## Maintainer rewrite

- Restrict MCP image writes to `ORACLE_HOME_DIR/generated` by default; explicit `ORACLE_MCP_ALLOW_EXTERNAL_OUTPUT=1` remains the operator opt-in for external paths.
- Reject traversal, escaping symlinks, dangling symlinks, Oracle config/session/browser-profile paths, and the generated directory itself as a file target.
- Fail closed for API-engine and remote-browser image output, including dry runs.
- Omit signed source/download URLs from structured MCP output; return local path, MIME type, size, dimensions, and file ID only.
- Preserve error output shape, reject known non-ChatGPT model families in `chatgpt_image`, and avoid sessionless image filename collisions.
- Update MCP/browser docs and changelog; changelog credits @umutkeltek.

Maintainer commits: `7dcad16e`, `34f9e9b9`, `b17aefa3`, `ea6fd28d`.

## Authenticated live proof

Real `chatgpt_image` MCP call through the existing signed-in Chrome profile:

- Session: `pr229-native-image-proof`
- Model: `gpt-5.5-instant`
- Status: completed
- Local artifact: `~/.oracle/generated/pr229-native-image-proof-20260611.png`
- PNG: 1254 x 1254, 786,988 bytes
- SHA-256: `a7ec5e6bd77fc1a28a6d2e61206176ad8a6bb580a11f22579218eaffdbb2aec9`
- Structured response returned `images[]` and `artifacts[]` with local metadata; no signed URLs were exposed.

## Verification

- `pnpm check`
- `pnpm test`: 1,213 passed, 41 skipped
- `pnpm docs:check`
- `pnpm docs:site`
- `pnpm build`
- `pnpm test:packed-cli`
- `pnpm test:mcp`
- `$autoreview --mode branch --base origin/main`: clean, no accepted/actionable findings


umutkeltek:
Did an independent pass on this and pushed a follow-up commit (`cfa4f047`) to address the review. Summary of findings + what changed.

## Findings

The bot's two P1s (output-path security boundary, needs real proof) are fair. Two refinements after reading the actual write path (`chatgptImages.ts`, `sessionRunner.ts`):

1. **`outputPath` is the stronger write primitive, not the image bytes.** The image saver partly self-limits — `resolveSiblingImagePath` forces a content-type extension on extensionless paths, and the bytes are ChatGPT-rendered image data (not attacker-chosen). But `outputPath` writes the model's **text** response verbatim (`sessionRunner.ts:991-1007`), text is fully caller-steerable, and its only guard is "not inside the sessions dir" — a state-corruption guard, not a security boundary. It `mkdir -p`s the parent, overwrites, and on `EPERM` even falls back to another path. Containment has to cover **both** fields.
2. **Default image path collides under parallelism (bot missed this).** `resolveDefaultImageOutputPath` was `chatgpt-image-${Date.now().toString(36)}.png`. Two concurrent `chatgpt_image` calls in the same millisecond resolve to the same path and overwrite each other — wrong default for a tool built for parallel agents.
3. **Output schema didn't declare its advertised shape.** `images/artifacts/resolved` were `z.record(z.string(), z.any())`, so the "returns saved paths in `structuredContent.images`" contract wasn't actually typed.

## Changes in `cfa4f047`

- **Containment (both fields):** `resolveMcpOutputPath` constrains agent-supplied `generateImage` / `outputPath` to `ORACLE_HOME_DIR` by default; `..` traversal is rejected by the resolved-prefix check. Opt out with `ORACLE_MCP_ALLOW_EXTERNAL_OUTPUT=1`. Rejected paths return a clean MCP error (not a throw). CLI `--generate-image` / `--output` are untouched — the boundary is MCP-only.
- **Collision fix:** default image path (and the browser-side session-less fallback) now carry a random suffix.
- **Contract:** `chatgpt_image` output reuses the typed `consult` output shape (`images` / `artifacts` / `resolved`) so the structuredContent shape is consistent and declared.
- **Tests (the bot's P1 rank-up move):** allow/reject path coverage for `mapConsultToRunOptions`, the `consult` MCP handler, and the `chatgpt_image` handler, plus a default-path-uniqueness assertion. Full MCP + image suites green (43 tests); format/lint clean.

## Still open

- **Residual symlink escape:** `path.resolve` doesn't resolve symlinks, so a symlinked dir under `ORACLE_HOME_DIR` pointing outside could bypass the prefix check. Left as a known follow-up (would need `realpath` on the existing parent).
- **Live proof:** the cheap `dryRun:true` path now exercises resolution + containment in tests, but a single redacted real ChatGPT image run (cookie download → populated `structuredContent.images[]`) is still worth attaching before merge — automation can't supply that.

umutkeltek:
Both remaining P1s addressed.

**Symlink hardening (`4ee828de`)** — `resolveMcpOutputPath` no longer relies on a lexical prefix check. It resolves symlinks in the existing path prefix via `realpath` (deepest existing ancestor + the not-yet-created remainder) and compares against the realpath of the Oracle home, so a symlinked parent under `ORACLE_HOME_DIR` (e.g. `generated -> /tmp/evil`) can't smuggle a write outside the boundary. Regression tests cover symlink escape for **both** `generateImage` and `outputPath` (rejected by default), a within-home symlink (allowed), and an escape allowed under `ORACLE_MCP_ALLOW_EXTERNAL_OUTPUT=1`. Verified the escape tests fail against the prior lexical check and pass after.

**Live behavior proof** — ran a real `chatgpt_image` MCP call end-to-end (redacted output in the PR body): GPT-5.5 browser session → image generated → cookie-backed download → saved under `ORACLE_HOME_DIR/generated/` with a populated `structuredContent.images[]` (`path`, `mimeType`, `sizeBytes: 962854`, `1254x1254`, `fileId`). File verified on disk as `PNG image data, 1254 x 1254`.

`tests/mcp` 41 passed; `oxfmt`/`oxlint` clean on the listed files; PR body updated with details.

@clawsweeper re-review

umutkeltek:
No code changes since the last review cleared the security defect (paths confined under the realpathed Oracle home, symlink escapes tested, external writes gated behind `ORACLE_MCP_ALLOW_EXTERNAL_OUTPUT`), with live `chatgpt_image` proof in the body. Requesting a fresh pass to confirm the current verdict.

@clawsweeper re-review

umutkeltek:
Closed the remote image-output [P1] from the last pass.

**`8f221a7a`** — image output now **fails closed** when a remote browser service is configured (`ORACLE_REMOTE_HOST`): the remote executor doesn't transfer image artifacts back, so rather than silently returning empty `structuredContent.images`, the run is rejected with a clear error pointing at the local-browser path. Local image generation is unchanged (proven live in the body). Regression test added (remote host + token + `generateImage` → `isError`, verified to fail against the pre-fix source); documented in `docs/mcp.md` + changelog. Full MCP suite 42 passed; format/lint/typecheck clean.

@clawsweeper re-review

PR #232, opened 2026-05-30
fix(browser): extract Deep Research report from out-of-process iframe

### LINKED ISSUE #244: Browser mode does not persist generic ChatGPT downloadable file artifacts
## Summary

When using Oracle's browser engine with ChatGPT, the assistant can return downloadable file artifacts such as source ZIPs, wheels, and sdists. Oracle recognizes these links well enough to include them in the generated markdown transcript, but it does not download/persist the actual files as session artifacts.

This seems adjacent to / a natural extension of #169, which added durable browser session artifacts (`transcript.md`, Deep Research reports, and generated-image downloads). The current artifact path appears to handle images and reports, but not generic ChatGPT downloadable files.

## Observed behavior

Oracle browser run against ChatGPT `gpt-5.5-pro` returned output like:

```markdown
Done — I created a minimal `uv`-based `llm-tool-bx` plugin project.

Files:

* [Source zip](sandbox:/mnt/data/llm-tool-bx-source.zip)
* [Wheel](sandbox:/mnt/data/llm-tool-bx/dist/llm_tool_bx-0.1.0-py3-none-any.whl)
* [sdist](sandbox:/mnt/data/llm-tool-bx/dist/llm_tool_bx-0.1.0.tar.gz)

I followed...
```

The markdown/transcript contains proper links for the generated file artifacts, so Oracle is seeing the assistant response and preserving the references. However, the linked files themselves are not downloaded into `~/.oracle/sessions/<id>/artifacts/` or surfaced as local file artifact paths in CLI output.

## Expected behavior

For ChatGPT browser sessions, Oracle should detect generic downloadable artifacts in assistant responses and persist them, similar to the existing image/report artifact handling from #169.

For example, after completion / harvest / render:

- `~/.oracle/sessions/<id>/artifacts/llm-tool-bx-source.zip`
- `~/.oracle/sessions/<id>/artifacts/llm_tool_bx-0.1.0-py3-none-any.whl`
- `~/.oracle/sessions/<id>/artifacts/llm_tool_bx-0.1.0.tar.gz`

and `oracle session <id> --render` / structured output would list them as saved artifacts.

## Environment

- Oracle CLI: `0.13.0`
- Engine: browser, driving `chatgpt.com`
- Model: ChatGPT `gpt-5.5-pro`
- Browser: Google Chrome Beta `150.0.7871.4` (Official Build) beta (arm64)
- OS: macOS arm64

## Notes / ask

@steipete could you give guidance on the preferred direction here?

I can prepare a PR if you point me at the intended design boundary. My current guess is that this should extend the browser artifact layer introduced in #169: detect non-image downloadable ChatGPT artifacts from the assistant turn / rendered markdown, fetch them using the live browser session cookies, save them under the session artifacts directory, and expose them through session render/status metadata.

Questions:

1. Should generic downloads be limited to ChatGPT `sandbox:/mnt/data/...` links, or should Oracle also inspect ChatGPT backend download URLs / DOM anchors directly?
2. Should this reuse the existing generated-image cookie-backed download path, generalized to arbitrary artifact links?
3. Any preferred artifact metadata shape for generic files (`kind: "file"`, mime type, size, original sandbox URL, local path)?

umutkeltek:
## Problem

Deep Research runs via the browser engine stopped returning results. The session hangs until timeout (or stays stuck `running`), and harvest captures only the `"ChatGPT said:"` placeholder instead of the report.

## Root cause

ChatGPT moved the Deep Research report out of the inline assistant turn and into a **doubly-nested, out-of-process sandboxed iframe**:

```
chatgpt.com/c/…                            → assistantCount: 0  (no report in the main DOM)
 └─ connector_openai_deep_research (OOPIF)  → body empty (sandbox shell)
     └─ nested same-origin iframe           → the actual report ("Research completed in …")
```

`waitForDeepResearchCompletion` has two extraction paths:

- `readDeepResearchFrameResult` — reads the **main page's** frame tree via isolated worlds. The OOPIF does **not** appear in the main page's frame tree, so `findDeepResearchFrameId` returns `null`. It can never see the report.
- `readDeepResearchTargetResult` — attaches to the iframe's own CDP target and walks its nested frames. This one **can** read the report.

The dispatch was `Page ? frameResult : client ? targetResult : null`. In production **both `Page` and `client` are always passed** (see `index.ts` and `reattach.ts`), so it always took the broken frame path and the working target path was effectively dead code. The completion poll never saw `finished` (no assistant turn in the main DOM either), so the run timed out.

A second guard (`hasActiveScopedResearch`) also depended on a main-DOM assistant turn that no longer exists, which would have blocked the report even after fixing dispatch.

## Fix

- Prefer the target-attach path (reaches the OOPIF + nested frames); fall back to the in-page frame path for legacy/inline rendering.
- Treat a **target-confirmed** completion as authoritative even when the main DOM exposes no assistant turn, since the report now lives entirely in the OOPIF.

The scoped-staleness guard is preserved for the in-page frame path (target-confirmed completions read the live connector iframe directly, so they don't need the main-DOM corroboration).

## Verification

Reattaching to a real completed Deep Research session now works end-to-end through the actual CLI:

```
Monitoring Deep Research (timeout: 40min)...
Deep Research completed (6s elapsed)
[browser] Saved deep-research-report artifact … deep-research-report.md (26.7 KB)
Reattach succeeded; session marked completed.
```

Previously this same flow extracted only `"ChatGPT said:"` and ran to timeout.

## Tests

- Adds a regression test: production shape (both `Page` and `client` passed, scoped run, `minTurnIndex >= 0`) where the report is only reachable via the target path and the main DOM has no assistant turn → returns the report instead of hanging.
- Existing staleness-protection test (`does not complete from an unscoped frame result during a scoped run`) still passes — the guard relaxation only applies to target-confirmed completions.
- `deepResearch.test.ts`: 28 passed. Full browser suite: 399 passed / 1 skipped.

## Known follow-up (not in this PR)

`readDeepResearchTargetResult` enumerates targets browser-wide (`Target.getTargets`). With multiple ChatGPT tabs open in the persistent profile, each holding a completed Deep Research report, a run could in principle pick up another tab's report. A follow-up should scope target enumeration to the current page session (the `client` is already connected to the page target) and drop the browser-wide scan.

---

## Follow-up: cross-tab scoping + wrapper-path binding + fresh live proof

The first review flagged that target-first extraction scanned browser-wide CDP targets (cross-tab leak). Addressed in two commits:

- **`53b7b331`** — scope discovery to the current page via page-session auto-attach; drop the browser-wide `Target.getTargets`/`attachToTarget` scan.
- **`95903b61`** — the page-scoping only held for direct-tab clients. On the browser-WSEndpoint path, `client` is a session-bound wrapper (`createSessionBoundChromeClient`) whose raw `send` is browser-level, so `Target.setAutoAttach` via `send` still went browser-wide there. Now the wrapper is tagged with its page session id (`oraclePageSessionId`) and auto-attach is bound to it explicitly. The cross-tab boundary now holds on both connection paths.

**Regression tests** (both verified to FAIL against the pre-fix source):
- foreign completed Deep Research target is never read (page-scoped discovery)
- every `Target.setAutoAttach` on a session-bound wrapper client is bound to the page session

**Fresh live multi-tab proof (latest head `95903b61`)** — a signed-in Chrome with **two** completed Deep Research tabs open; the real `waitForDeepResearchCompletion` run scoped to each tab returns that tab's own report, no cross-contamination:

```jsonc
{
  "tabA": { "name": "v2 (Routing Cost Map)",  "len": 27099,
            "head": "Current LLM Inference Routing Cost Map / Scope and reading notes ...",
            "hasOwn_RoutingMap": true,  "leaked_fromB": false },
  "tabB": { "name": "v1 (Provider Cost Map)", "len": 14317,
            "head": "LLM Inference Provider Cost Map / Scope and normalization ...",
            "hasOwn_ProviderMap": true, "leaked_fromA": false }
}
// ISOLATION PASS — each tab returned its OWN report, no cross-contamination
```

Both connector OOPIFs (`connector_openai_deep_research.*`) were present simultaneously; tab A extracted the Routing report and tab B the Provider report, each from its own page-scoped session.

**Verification**: `tests/browser/deepResearch.test.ts` 30 passed; full browser suite 401 passed / 1 skipped; typecheck (changed files) / format / lint clean.

---

## Update — candidate-order fix, clean-up, and re-proof at HEAD (`f5428df9`)

Addressing the re-review of the page-session-binding commit:

**Correctness (`70375b47`, `8a9b98c2`)**
- A page can expose more than one Deep Research iframe target. Target scanning now returns a completed read immediately and only keeps the best in-progress/text-bearing read when none completed — it no longer returns the first in-progress target and misses a later completed OOPIF.
- `waitForDeepResearchCompletion` no longer lets an incomplete target read suppress the in-page frame fallback; a completed target read stays authoritative, otherwise the frame path runs (preserving legacy inline/mixed rendering — the compatibility concern).

**Clean-up (`f5428df9`)**
- Extracted the read-selection logic into a documented, unit-tested helper `pickPreferredDeepResearchRead`; renamed the misleading `frameResult` local to `read`.

**Tests** (each verified to fail against the pre-fix source):
- first-target-in-progress + later-target-completed → returns the completed report
- in-progress target + completed in-page frame → returns the frame report (legacy inline)
- 6 direct unit tests for the selection helper, including the legacy `no-target + completed-frame` case
- `tests/browser/deepResearch.test.ts`: all pass; full browser suite 410 passed / 1 skipped; typecheck (changed files) / format / lint clean.

**Fresh live multi-tab proof at HEAD `f5428df9`** — signed-in Chrome, two completed Deep Research tabs open simultaneously (both connector OOPIFs present); the real `waitForDeepResearchCompletion` scoped to each tab returns that tab's own report:

```jsonc
{
  "head": "f5428df9",
  "tabA": { "name": "v2 (Routing Cost Map)",  "len": 27099,
            "head": "Current LLM Inference Routing Cost Map / Scope and reading notes ...",
            "hasOwn_RoutingMap": true,  "leaked_fromB": false },
  "tabB": { "name": "v1 (Provider Cost Map)", "len": 14317,
            "head": "LLM Inference Provider Cost Map / Scope and normalization ...",
            "hasOwn_ProviderMap": true, "leaked_fromA": false }
}
// ISOLATION PASS — each tab returned its OWN report, no cross-contamination
```

---

## Maintainer rewrite and final validation (`9097283f`)

This section supersedes the earlier "known follow-up" note. The contributor's six commits remain intact and credited; three maintainer commits harden the target boundary and current-run selection:

- `a20b13b6` filters flattened `Target.attachedToTarget` events by their parent page session, confirms Deep Research targets before evaluating them, and adds the foreign-page-session regression.
- `a3c7d39d` snapshots confirmed target IDs immediately before submission so an earlier completed report in the same conversation cannot satisfy the new run.
- `9097283f` makes reattach fail closed against stale reports by mapping each OOPIF root frame through `DOM.getFrameOwner` to its owning ChatGPT conversation turn and enforcing `minTurnIndex`. Selection does not depend on target order, target count, or report content equality.

The final implementation also preserves completed-target precedence, ignores later empty/in-progress targets when a valid completion exists, and keeps the legacy in-page frame fallback.

### Authenticated live proof

Validated in the existing signed-in Chrome profile against two distinct completed Deep Research conversations. Sensitive report content and connector details were not printed.

- Main DOM: no assistant report turns or completion label; report exists only in the sandboxed connector OOPIF.
- OOPIF owner: the live iframe resolves to a concrete `conversation-turn-N` section (`conversation-turn-2` in the inspected conversation), confirming the reattach scoping signal used by `9097283f`.
- Conversation A cross-process snapshot: 20,200 bytes; SHA-256 `701f43cd382c08c8b6a4fd8d1a0001821cb3becbb52c66f682708e4ee3184599`; one completion marker; 41 URL-like sources.
- Conversation B cross-process snapshot: 13,739 bytes; SHA-256 `734c107cc5f50705472974c56ceee1ec8ee325bc624e52987a0631d585ce1972`; one completion marker; 39 URL-like sources.
- Distinct hashes confirm two different reports were read from their own page-scoped OOPIFs.

### Final gates

- `pnpm test`: 137 files passed, 18 skipped; 1,190 tests passed, 41 skipped.
- `pnpm check`: clean.
- `pnpm docs:check`: clean (85 flags, 6 files).
- `pnpm build`: clean.
- `pnpm docs:site`: clean.
- `pnpm test:packed-cli`: clean.
- Focused browser/reattach regressions: 130 tests passed.
- `$autoreview --mode local`: clean.
- `$autoreview --mode branch --base origin/main`: clean.

Fixes #244.


umutkeltek:
Addressed the P1 (browser-wide target scan / cross-tab leak) in `53b7b331`.

## Fix
`readDeepResearchTargetResult` no longer enumerates `Target.getTargets` (browser-wide). `client` is connected to the conversation **page** target, so enabling `Target.setAutoAttach` on that session only surfaces **this page's** related targets (its Deep Research OOPIF subframe). Discovery is now limited to those auto-attached, page-scoped sessions; the browser-wide `getTargets`/`attachToTarget` enumeration (and the now-unneeded `setDiscoverTargets`) is removed. A foreign completed Deep Research tab in a shared/persistent profile can no longer be read into the current session.

## Regression test
Added `does not return a foreign completed Deep Research report from another tab`: the current page's OOPIF auto-attaches but is still **in progress**, while a **foreign completed** report is visible via the browser-wide `Target.getTargets` scan. The test asserts the foreign report is never read (the run times out) and that the browser-wide scan is not consulted.

Proof it's a real guard: against the pre-scoping source the test **fails** (`foreignAttachCalled === true` — the foreign target was attached and its report returned); with the scoping fix it **passes**.

- `tests/browser/deepResearch.test.ts`: 29 passed. Full browser suite: 400 passed / 1 skipped. Typecheck (changed file) / format / lint clean.
- Changelog updated with the OOPIF fix + the cross-tab scoping.

@clawsweeper re-review

umutkeltek:
Both follow-up P1s resolved + fresh live proof on the latest head.

**Wrapper-path binding (`95903b61`)** — the prior page-scoping only held for direct-tab clients. On the browser-WSEndpoint path the session-bound wrapper's raw `send` is browser-level, so `Target.setAutoAttach` still went browser-wide. Now the wrapper carries its page session id (`oraclePageSessionId`) and auto-attach is bound to it explicitly. New regression test asserts every `setAutoAttach` on a session-bound wrapper is page-bound — verified to fail against the un-bound source.

**Fresh live multi-tab proof (latest head)** — signed-in Chrome, **two** completed Deep Research tabs open simultaneously (both `connector_openai_deep_research` OOPIFs present). The real `waitForDeepResearchCompletion` scoped to each tab returned that tab's own report:
- tab A → "Current LLM Inference **Routing** Cost Map" (27,099 chars), no leak from B
- tab B → "LLM Inference **Provider** Cost Map" (14,317 chars), no leak from A
- **ISOLATION PASS** — no cross-contamination.

Redacted result + commit details in the PR body. `tests/browser/deepResearch.test.ts` 30 passed; full browser suite 401 passed.

@clawsweeper re-review

umutkeltek:
Addressed the re-review of the page-session-binding commit, plus a clean-up pass and fresh proof at the new HEAD.

**Correctness (`70375b47`, `8a9b98c2`)** — target scanning now returns a completed read immediately and only keeps the best in-progress read when none completed (no longer returns the first in-progress target and misses a later completed OOPIF); and an incomplete target read no longer suppresses the in-page frame fallback (a completed target read stays authoritative — this preserves the legacy inline/mixed rendering path, the compatibility concern).

**Clean-up (`f5428df9`)** — extracted the read-selection into a documented, unit-tested helper (`pickPreferredDeepResearchRead`); renamed the misleading `frameResult` → `read`.

**Tests** (each proven to fail against the pre-fix source): first-target-in-progress + later-target-completed; in-progress target + completed in-page frame (legacy inline); 6 unit tests for the selection helper incl. the `no-target + completed-frame` legacy case. `tests/browser/deepResearch.test.ts` all pass; full browser suite 410 passed; typecheck (changed files) / format / lint clean. An independent code-review pass found nothing substantive.

**Fresh live multi-tab proof at HEAD `f5428df9`** (in the PR body): signed-in Chrome, two completed Deep Research tabs open at once, each scoped extraction returns that tab's own report — ISOLATION PASS, no cross-contamination.

@clawsweeper re-review

umutkeltek:
I was kinda obsessed with the rating, it would be very logical to gamify pr's like this in the future.

yeqiuqiu123:
Downstream validation from a real Hermes/Oracle autonomous Deep Research workflow:

- Reproduced on released `@steipete/oracle` v0.13.0: completed ChatGPT Deep Research sessions harvested successfully at the CLI transport layer (`rc=0`, state completed) but output only the placeholder `ChatGPT said:` / 13 bytes.
- Built and ran this PR head locally (`f5428df94497b354d9646d4cc8634e070c59d0d7`) as a profile-local hotfix.
- Re-harvested a known completed Deep Research session through `oracle session <slug> --harvest --write-output ...` and the same workflow produced real report content instead of the 13-byte placeholder.
- End-to-end runner gate also passed: the downstream harvester classified the patched output as completed/valid rather than garbage/empty.

Minimal observed before/after:

```text
before v0.13.0:
  rc=0
  state=completed
  output_size=13
  output="ChatGPT said:"

after PR head f5428df9:
  rc=0
  state=completed
  output_size=1476
  output contains expected Deep Research marker/content
```

This is not just a unit-test-only issue for us; it was blocking an autonomous Research Intern pipeline because completed reports were being discarded as empty extraction artifacts. The PR head fixes the real downstream failure mode.

PR #237, opened 2026-06-02
feat(browser): resume a ChatGPT thread via --followup (+ reattach guidance on failure)

### LINKED ISSUE #236: Bug: browser-engine follow-ups can't resume a prior ChatGPT thread (--followup is API-only, --browser-follow-up is launch-only)
## Summary

There's no way to continue an existing **browser (ChatGPT) conversation** across CLI invocations. `--followup` is API-only, and `--browser-follow-up` only queues turns *within the launching run* — so once a browser session completes, you can't ask it a follow-up that lands in the same thread (with its already-loaded context).

## Current behavior

- `oracle --followup <browser-session-id> -p "..."` → `✖ --followup requires --engine api`
- `--browser-follow-up "<prompt>"` is only consumed at launch (queued after the initial answer in the same process). It can't target a prior/stored session.
- The conversation URL *is* persisted (`browser.runtime.tabUrl` / `conversationId`), and `reattach.ts` already reopens it for `--harvest`/`--live` — but that path is read-only; there's no "submit a new turn into the reopened thread."

Net effect: to ask GPT‑5.5 Pro a quick follow-up after a big browser consult, you have to fire a brand-new run and **re-upload the entire packet**, paying the full token/time cost again, instead of adding one cheap turn to the warm thread.

## Expected

`oracle --followup <browser-session-id> -p "..."` should, when the referenced session is a browser run, **reopen that session's ChatGPT conversation and submit the prompt as the next turn** (attachments included), leaving API followups (`resp_…` / stored API sessions) unchanged. The model picker should be skipped on resume (the thread already has its model).

## Why it matters

A browser Pro Extended consult can carry hundreds of thousands of tokens of loaded context. A follow-up that reuses that thread is ~tens of tokens; re-running from scratch is the whole packet again (minutes + cost). Cross-run browser follow-ups make iterative "challenge / drill-down" workflows practical.

## Discovered sub-bug: composer wiped by late hydration on resume

While implementing this, a real race showed up that's worth flagging on its own (related to the closed #175 "stale composer text"): when you navigate to an **existing** conversation, the composer briefly reports "ready," the prompt gets typed — then the thread's prior history finishes hydrating, React re-renders, and the freshly-typed prompt is **wiped**. A fresh chat never hits this (no history to load); a large resumed thread reliably does.

Fix that worked: after navigating to the resumed conversation, wait for the prior turns to render *and stop growing*, let React settle, then re-confirm the composer before typing — mirroring the existing `recoverConversation` hydration-delay pattern.

## Bonus: reattach guidance on failure

Adjacent ergonomics: when a browser run fails/times out, oracle should print the exact reattach commands (`--render` / `--live` / `--harvest`) with the real session slug, so a long Pro run that completed in the background after a wall-clock timeout is never silently lost.

## Environment

- `@steipete/oracle` 0.13.0, browser engine (manual-login profile), macOS, Node 26.

## Note

I have a working fix for all three (browser `--followup` resume, the hydration race, and the reattach guidance), verified live end-to-end (resumed a real thread, prompt stuck, submitted to the same `/c/` id, no re-upload). Happy to open a PR — filing this first for tracking.

hbruceweaver:
Fixes #236.

## Contract

Extend the existing `--followup` contract based on the stored parent session:

- Browser parent: reopen the exact saved ChatGPT conversation and submit the prompt/files as the next turn.
- OpenAI/Azure parent or `resp_...`: preserve the existing Responses API continuation path.
- Planned turns in one browser run remain `--browser-follow-up`.

Browser resume inherits the parent's browser profile, configuration, and model. It clears stale tab targeting, skips the model picker, disables Deep Research for the resumed turn, and leaves the conversation unarchived.

## Safety

Resume fails closed before typing or submitting unless:

- the saved URL is an HTTPS `chatgpt.com` / `chat.openai.com` conversation URL on the default port;
- prior conversation turns hydrate and stabilize;
- the loaded conversation ID matches the saved conversation ID.

This removes the previous fallback-to-home behavior that could turn a failed resume into an unrelated fresh chat. Reattach guidance now prints only after a prompt was submitted or a recoverable conversation exists, and not for pre-conversation or Cloudflare failures.

## Implementation

- Resolve browser parents separately while preserving API follow-up behavior.
- Inherit stored browser configuration/model and persist browser parent lineage without requiring a Responses API ID.
- Verify resumed-thread hydration and identity in local, attached-tab, and remote browser paths.
- Add focused regression coverage for unsafe URLs, zero-turn hydration, wrong-thread redirects, browser lineage, and reattach guidance.
- Update CLI help, README, browser/session/configuration/follow-up docs, and changelog.

## Proof

- `pnpm run check`
- `pnpm run docs:check`
- `pnpm run build`
- `pnpm test`: 1,106 passed, 41 skipped
- Focused regressions: 95 passed, 1 skipped
- Structured autoreview: clean for both the maintainer patch and the full branch diff against `origin/main`
- Authenticated end-to-end browser proof:
  - resumed a completed saved ChatGPT browser session;
  - loaded four stable prior turns;
  - skipped the model picker;
  - submitted only after exact-thread verification;
  - parent and child conversation IDs had the same redacted SHA-256 prefix;
  - child lineage points to the parent;
  - inherited `gpt-5.5`, Chrome profile, and model label;
  - persisted exact response `FOLLOWUP_237_READY`;
  - completed in 13.8s.

Conversation IDs and authenticated browser details are intentionally redacted.

## Credits

Original implementation by @hbruceweaver. The maintainer rewrite incorporates fail-closed/configuration/lineage ideas from #243 by @pdurlej; both are credited in the changelog, and Piotr is co-authored on the maintainer commit.


hbruceweaver:
Addressed both review findings in 185e4aba.

**① Reuse the recoverable ChatGPT URL gate** (`src/cli/followup.ts`)
`resolveBrowserResumeConversationUrl` now resolves through the existing `resolveRecoveryUrl` (which prefers `browser.harvest.url` over `browser.runtime.tabUrl` and gates both with `isRecoverableChatGptConversationUrl`). Only when neither candidate is a recoverable `chatgpt.com/c/<id>` URL does it rebuild from a stored `conversationId` against the session's ChatGPT base — and that rebuilt URL is gated too. Home / project-shell / external URLs can no longer drive the signed-in profile.

Regressions added in `tests/cli/followup.test.ts`: harvest-url preference over a stale runtime tab URL, external-URL rejection, project-shell rejection, and non-ChatGPT `conversationId` rejection.

**② Apply the hydration wait to the remote path** (`src/browser/index.ts`)
Extracted the resumed-thread hydration wait into a shared `waitForResumedConversationHydration` helper (`src/browser/actions/navigation.ts`) and called it on the remote browser path after navigation when `resumeConversationUrl` is set — so `--remote-chrome` follow-ups no longer type before prior turns finish hydrating. The local path now wraps the whole wait in `raceWithDisconnect`, so a dropped client aborts immediately instead of polling to the hydration deadline.

Verification: `tsgo --noEmit` (full tsconfig, incl. tests) clean, oxlint clean, and the named suites pass — `pnpm vitest run tests/cli/followup.test.ts tests/browser/recoverConversation.test.ts tests/browser/sessionRunner.test.ts tests/cli/sessionRunner.test.ts` (58 + 10 green).

@clawsweeper re-review

PR #257, opened 2026-06-15
fix(browser): treat /api/auth/session as login authority over DOM login CTAs

amadad:
## Problem

On current ChatGPT, browser-mode runs fail for **authenticated** sessions with:

```
ChatGPT session not detected. Login button detected on page.
```

Verbose probe output on a session that is genuinely logged in:

```
Login probe failed (sessionStatus=200, sessionAuthenticated=true,
sessionResolved=true, domLoginCta=true, appAuthenticated=true, onAuthPage=false)
```

Two ChatGPT-side changes cause this, neither fixed by re-login or cookie refresh:

1. **Login CTA on authenticated pages.** ChatGPT renders a visible "Log in" button (`data-testid="login-button"`) pinned to the sidebar nav's sticky bottom *even when fully authenticated*. `hasLoginCta()` matches it and vetoes the session, despite `/api/auth/session` returning an authenticated user and the app shell being authenticated.
2. **Device-bound sessions.** Sessions are now bound to a set of `oai-*` cookies. Without them, only ~4 cookies are copied, the app renders a half-authenticated shell, and `/backend-api/*` can return 401 for an otherwise valid session.

## Fix

- **Session-endpoint authority over DOM CTA.** A resolved authenticated `/api/auth/session` + authenticated app shell (and not on an auth page) is authoritative; a DOM login CTA no longer vetoes it. CTAs remain authoritative for the weaker fallback classifications, and auth-page redirects still fail closed.
- **Exclude sidebar-nav CTAs** (`node.closest('nav')`) from the login signal. Genuine logged-out pages place login CTAs in the main content.
- **Mid-run `validateChatGPTSession`** only treats a login CTA as a logout signal when the composer (textarea) is also gone.
- **Diagnostics:** the probe now reports `domLoginCtaDetail`, logged as `Login CTA matched: <element>`, so future drift is diagnosable from `--verbose`.
- **Device-binding cookies** added to `DEFAULT_CHATGPT_COOKIE_NAMES` (`oai-did`, `oai-sc`, `__Secure-oai-is`, …).

## Verification

Live against ChatGPT (browser serve mode, GPT-5.5 Pro):

- Before: stock 0.14.0 → `Login probe failed (... sessionAuthenticated=true ... domLoginCta=true)`, "Copied 4 cookies", run fails in ~32s.
- After: "Copied 13 cookies", no probe failure, run completes (`Answer: READY`, ~53s).
- `tests/browser/pageActions.test.ts` updated; 59 passing locally.

Happy to split the cookie-list change into its own commit or adjust the authority formula if you'd prefer a narrower fix.

### DIFF
--- a/src/browser/actions/navigation.ts
+++ b/src/browser/actions/navigation.ts
@@ -259,6 +259,9 @@ export async function ensureLoggedIn(
       probe.appAuthenticated,
     )}, cfBlocked=${Boolean(probe.cfBlocked)}, url=${probe.pageUrl ?? "n/a"}, error=${probe.error ?? "none"})`,
   );
+  if (probe.domLoginCtaDetail) {
+    logger(`Login CTA matched: ${probe.domLoginCtaDetail}`);
+  }
 
   const domLabel = probe.domLoginCta ? " Login button detected on page." : "";
   const cookieHint = options.remoteSession
@@ -638,6 +641,7 @@ type LoginProbeResult = {
   error?: string | null;
   pageUrl?: string | null;
   domLoginCta?: boolean;
+  domLoginCtaDetail?: string | null;
   onAuthPage?: boolean;
   appAuthenticated?: boolean;
   backendStatus?: number | null;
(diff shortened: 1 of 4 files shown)


amadad:
Closing — opened in error, not intended for upstream. Apologies for the noise.

PR #258, opened 2026-06-15
feat(browser): add --copy-profile for login-free reuse of a signed-in Chrome profile

edwarddgao:
## What

Adds `--copy-profile <dir>` to browser mode: copy a signed-in Chrome user-data directory to a throwaway profile and run against it, reusing your live ChatGPT session with **no manual sign-in**, then delete the copy.

```bash
oracle --copy-profile "$HOME/Library/Application Support/Google/Chrome" \
  --engine browser --model gpt-5.5-pro -p "review this plan" --file "src/**"
```

## Why

Browser mode today reaches a signed-in session two ways, each with friction: cookie-sync (Keychain prompt, and ChatGPT increasingly rejects transplanted cookies) or `--browser-manual-login` (interactive sign-in the server periodically invalidates). For users already logged into ChatGPT in their everyday Chrome, copying the whole profile to a throwaway dir and driving that is the most reliable login-free path — it reuses the real, device-bound session in place rather than transplanting cookies.

## How

- New `src/browser/profileCopy.ts`: `rsync -a` the source `Default/` (minus cache dirs) + top-level `Local State` into a temp dir. `Local State` is required — it holds the Keychain-wrapped "Chrome Safe Storage" key that decrypts the cookies; a cookies-only copy fails the login check. rsync tolerates a live source profile (exit 24).
- `--copy-profile` reuses the existing ephemeral temp-profile path in `runBrowserMode` (created via `mkdtemp`, deleted on teardown), seeded from the copy before launch. It composes with the existing "already signed in" handling: manual-login and copy-profile share one `profileIsPreSigned` guard that skips the cookie clear + sync.
- `launchChrome` skips the keychain-mocking flags for copy-profile runs (chrome-launcher's default `--use-mock-keychain`/`--password-store=basic` via `ignoreDefaultFlags`, plus Oracle's own) so the launched Chrome decrypts the copied profile's cookies with the real Keychain.
- `copyProfileSource` is threaded through the config types exactly like `manualLoginProfileDir`.

## Lifecycle / safety (addressed in review follow-up `99c4ed74`)

The copy contains a full signed-in profile, so it is treated as sensitive throwaway data:
- **Always deleted**, including on interrupt: copy-profile runs reject `--browser-keep-browser`, and the signal/in-flight teardown force-cleans the copied profile (`forceProfileCleanup`) so an interrupted run never leaves it on disk.
- **Fails fast** if the required `Local State` cannot be copied (no silent "looks logged-out" runs).

## Testing

- `pnpm run lint` (typecheck + oxlint): clean.
- `pnpm test`: full suite green (**1281 passed**); unit tests cover the config mapping, the `--browser-keep-browser` rejection, and the `Local State` fail-fast.
- Manual e2e (macOS, Chrome 149), redacted — shows login-free reuse **and** cleanup:

```
# profile temps before: 0
$ oracle --copy-profile "$HOME/Library/Application Support/Google/Chrome" \
    --engine browser --model gpt-5.5-pro --browser-model-strategy select -p "<task>"
[browser] Model selection evidence: requested=Pro; resolved=Pro Extended; status=already-selected; strategy=select; verified=yes.
Answer: OK.
14.0s · gpt-5.5-pro[browser]
# profile temps after: 0      <- copied throwaway profile created and deleted
```

(No sign-in prompt — it authenticated via the copied session. 0 temp profiles before and after, confirming cleanup. The model self-reports unreliably in the answer text; the `verified=yes` picker check is authoritative.)

## Caveats

- macOS/Linux only (Keychain-encrypted cookies; requires `rsync`). On macOS the launching Chrome must be the install that owns the "Chrome Safe Storage" Keychain item.
- Drives the ChatGPT web UI via automation — same supportability caveats as existing browser mode.


edwarddgao:
@clawsweeper re-review — addressed both findings in 99c4ed74: copy-profile now rejects `--browser-keep-browser` and force-cleans the copied profile on interrupt/in-flight (never left on disk), and fails fast if `Local State` can't be copied. Added tests for both. PR body now includes redacted real-browser proof of login-free reuse + 0→0 temp-profile cleanup.

steipete:
Maintainer verification complete on exact head `4324ebf212b6e61a2947d92966c81ab98f77ddd2`:

- Rebased onto current `main`; active/non-Default Chrome profiles are now selected from `Local State`, with explicit `--browser-chrome-profile` override coverage.
- Copied profiles are deleted after setup/launch failures, normal errors, incomplete captures, Cloudflare challenges, connection loss, and signals. Persistent, attach-running, and remote modes now fail closed instead of silently bypassing the copy.
- Focused final profile/lifecycle/config suites: 115/115 passed. Full suite: 1,336 passed, 43 skipped. Lint, typecheck, docs/help validation, skill frontmatter parsing, and build passed.
- Real Chrome E2E on the final branch copied and launched a local Chrome profile, correctly failed authentication for that logged-out source, removed every throwaway profile, and persisted no reattach runtime. A second CLI E2E confirmed configured remote-host routing rejects `--copy-profile` before dispatch. The contributor's authenticated Chrome E2E remains the positive login-reuse proof: no sign-in prompt and zero temporary profiles before/after.
- Autoreview was iterated through all actionable findings; final full-branch review is clean.
- Exact-head CI run 27821331304: Ubuntu, macOS, Windows, and GitGuardian all green.

This closes the retained-session-data finding and the additional active-profile, custom-debug-host keychain, conflicting-mode, and stale-reattach edge cases found during maintainer review.

PR #267, opened 2026-06-19
fix(browser): bound Instant model selection

### LINKED ISSUE #263: Browser model picker hangs with --browser-model-strategy select on current ChatGPT Pro UI
## What happened

Small but blocking one 🧿

I can reproduce a browser-mode hang with `--browser-model-strategy select` on the current ChatGPT Pro UI. The same remote Chrome session works with `--browser-model-strategy current`, but `select` opens/reaches the picker and then never returns cleanly.

This is related to #182, but not quite the same shape from what I can see: #182 is closed, while this still reproduces for me on `@steipete/oracle@0.14.1`.

## Repro

Environment:

- `@steipete/oracle@0.14.1`
- Node `24.x`
- Browser mode with remote Chrome DevTools
- Chrome `148.0.7778.178` via `selenium/standalone-chromium:4.44.0-20260505`
- Signed-in ChatGPT Pro account
- Current visible composer model pill: `Pro Extended`

This works, but of course uses the currently selected model:

```bash
oracle \
  --engine browser \
  --model gpt-5.5-instant \
  --browser-model-strategy current \
  --browser-archive never \
  --timeout 15m \
  --write-output /tmp/oracle-current.md \
  --browser-tab current \
  --remote-chrome <host:port> \
  -p "Reply exactly oracle-current-ok."
```

This hangs:

```bash
oracle \
  --engine browser \
  --model gpt-5.5-instant \
  --browser-model-strategy select \
  --browser-archive never \
  --timeout 15m \
  --write-output /tmp/oracle-select.md \
  --browser-tab current \
  --remote-chrome <host:port> \
  -p "Reply exactly oracle-select-ok."
```

I also reproduced it by calling `ensureModelSelection(Runtime, "gpt-5.5-instant", logger, "select")` directly through CDP; it timed out after 35s.

## What I see in the DOM

When the picker is open, the relevant bits look like this:

```text
model button: button.__composer-pill
model text: Pro Extended
visible menu item: role="menuitem", text="GPT-5.5"
active element text: GPT-5.5
```

For the exact current tab, `current` strategy succeeds and logs model selection evidence roughly like:

```text
requested=GPT-5.5 Instant; resolved=Pro Extended; status=already-selected; strategy=current; verified=no
```

That is fine for the workaround, but it is not safe for a gateway to expose a route named `gpt-5.5-instant`, because it would silently run the already-selected Pro model.

## Expected

`select` should either:

1. select the requested model/variant and continue, or
2. fail fast with a clear `option-not-found` / unsupported-model error if the current ChatGPT UI no longer exposes the requested variant in that picker state.

The bad state is hanging with the picker open and no output file written.

## Why this matters

I am using Oracle behind an OpenAI-compatible internal gateway. Because `select` is unreliable here, I can only expose the already-selected Pro route using `--browser-model-strategy current`. Any other model-named route would be misleading.

Happy to test a patch or a canary release against the same remote-Chrome setup.

steipete:
Fixes #263.

## Changes

- rank concrete GPT-5.5 Instant above generic GPT-5.5 submenu rows
- prevent reopening the same submenu after React replaces its DOM node
- enforce the model-selection deadline across retries
- add live-shaped regression coverage for successful and missing Instant options

## Proof

- `pnpm test` — 1,300 passed, 43 skipped
- format, typecheck, lint, and focused model-selection tests passed
- autoreview: no accepted/actionable findings
- live authenticated browser E2E: requested `gpt-5.5-instant` with strategy `select`; picker reported verified `already-selected`; ChatGPT returned exactly `ORACLE_ISSUE_263_INSTANT_SELECT_OK`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Browser: wait for the current ChatGPT Intelligence pill before falling back to the default thinking level. This lets `--browser-model-strategy current --browser-thinking-time extra-high` select and verify Extra High before submitting (thanks @alex-on-java).
+- Browser: wait for the current ChatGPT Intelligence pill before falling back to the default thinking level, and make `--browser-model-strategy select` prefer concrete requested variants over version-only submenu wrappers with bounded retries. This lets current-model runs select and verify Extra High before submitting and prevents explicit Instant selection from hanging (thanks @alex-on-java and @servrox).
 
 ## 0.14.1 — 2026-06-15
 
--- a/src/browser/actions/modelSelection.ts
+++ b/src/browser/actions/modelSelection.ts
@@ -711,10 +711,27 @@ function buildModelSelectionExpression(
         normalizedTestId.includes('pro');
       const candidateHasInstant =
         normalizedText.includes('instant') || normalizedTestId.includes('instant');
+      const candidateSelectsConfiguredVersion =
+        Boolean(getConfigurationDialog()) &&
+        candidateSelectsDesiredVersion &&
+        (node?.getAttribute?.('role') === 'option' ||
+          node?.getAttribute?.('role') === 'menuitemradio');
+      const candidateOpensInstantSubmenu =
+        wantsInstant &&
+        candidateSelectsDesiredVersion &&
+        !candidateHasInstant &&
+        (normalizedTestId.includes('submenu') ||
+          node?.getAttribute?.('aria-haspopup') === 'menu' ||
+          node?.getAttribute?.('data-has-submenu') !== null);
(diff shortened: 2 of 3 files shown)


steipete:
Landed as `f73dce42`.

Proof:

- full suite: 1,300 passed, 43 skipped
- focused model-selection regressions, format, typecheck, lint, and autoreview passed
- live authenticated browser E2E requested `gpt-5.5-instant` with strategy `select`; selection was verified and ChatGPT returned exactly `ORACLE_ISSUE_263_INSTANT_SELECT_OK`
- exact-head CI run 27815735403 is green on Ubuntu, macOS, Windows, and GitGuardian

PR #274, opened 2026-06-25
feat(config): add user model overrides for custom gateways

### LINKED ISSUE #273: Feature: user-config per-model overrides (apiModel + reasoning effort) for custom OpenAI-compatible --base-url endpoints
### Summary

When running against a **custom OpenAI-compatible gateway** via `--base-url` / `apiBaseUrl` (e.g. a LiteLLM proxy or a self-hosted gateway), there is currently **no supported way to override a model's `apiModel` (the on-wire model id) and `reasoning` effort from user config**. The only way to make such an endpoint work correctly is to hand-edit the bundled `MODEL_CONFIGS` in the published package — which is wiped on every reinstall/upgrade.

This is a gap specifically for the `--base-url` path that the CLI already advertises ("Override the OpenAI-compatible base URL for API runs (e.g. LiteLLM proxy endpoint)").

### Motivation / real-world setup

Many OpenAI-compatible gateways **rename model ids** in their catalog. For example, a gateway may expose `gpt-5.5` as `gpt-5.5-mygateway` (vendor suffix), and the model may require a non-default `reasoning.effort` (e.g. `high`/`xhigh`) to think hard.

Two things break today:

1. **Passing the gateway's renamed id directly** (`-m gpt-5.5-mygateway`) hits the generic synthesized config in `resolveModelConfig` (`src/oracle/modelResolver.ts`), which sets `reasoning: null`, `inputLimit: 200_000`, and a GPT-5-pro tokenizer. So reasoning effort is **silently dropped**, the context limit is wrong, and token estimates are off.
2. **Passing the canonical id** (`-m gpt-5.5`) preserves `reasoning`/tokenizer/`inputLimit` from `MODEL_CONFIGS`, but then the request is sent with the **bundled `apiModel`** (`model: modelConfig.apiModel ?? modelConfig.model` in `src/oracle/request.ts`), which the gateway doesn't recognize.

There is no way to get *both* the renamed `apiModel` *and* the correct `reasoning`/metadata without editing bundled code.

### Why user config can't do it today

- User config (`~/.oracle/config.json`, `src/config.ts`) accepts `engine`, `model`, `apiBaseUrl`, `search`, `maxFileSizeBytes`, `notify`, `heartbeatSeconds`, `filesReport`, `background`, `promptSuffix`, and a `browser{}` block. There is **no per-model override field**.
- Project config is sanitized to a fixed allowlist (`sanitizeProjectConfig` in `src/config.ts`) that intentionally excludes model routing — good, that should stay.
- `resolveModelConfig` already composes config from multiple sources: known `MODEL_CONFIGS`, OpenRouter catalog enrichment (`isOpenRouterBaseUrl` path), and a generic fallback. **A user-config override layer would be a natural additional source**, mirroring how OpenRouter enrichment already augments/overrides fields like `apiModel`, `inputLimit`, and `pricing`.

### Proposed solution

Allow a **`modelOverrides` map in user config only** (not project config, for safety) that merges over known `MODEL_CONFIGS` entries inside `resolveModelConfig`:

```json5
{
  "apiBaseUrl": "https://my-gateway.example/v1",
  "model": "gpt-5.5",
  "modelOverrides": {
    "gpt-5.5": {
      "apiModel": "gpt-5.5-mygateway",
      "reasoning": { "effort": "xhigh" },
      "inputLimit": 1050000,
      "pricing": { "inputPerToken": 0.000005, "outputPerToken": 0.00003 }
    }
  }
}
```

Then `oracle -m gpt-5.5` (or `model` in config) routes to the gateway's id while keeping correct reasoning/tokenizer/limits.

### Suggested scope for a first PR (intentionally tight)

- **User config only** (skip project config — keeps the existing sanitize allowlist boundary intact).
- **Override existing known model keys only** (inherit tokenizer from `MODEL_CONFIGS` — avoids the "JSON can't express a tokenizer function" problem). New custom aliases / tokenizer selection / `extends` inheritance can be a follow-up.
- Whitelist a small set of overridable fields: `apiModel`, `reasoning`, `inputLimit`, `pricing`.
- Apply the override as the last layer in `resolveModelConfig`.
- Support `reasoning: null` to **explicitly clear** a known model's effort (don't blind deep-merge in a way that prevents clearing).
- No new CLI flags in the first cut (a per-model map avoids the multi-model ambiguity that `--api-model` would have with `--models a,b`).

### Risks / open questions

- **Provider-specific reasoning shape**: `reasoning` is forwarded as-is to the request body (`src/oracle/request.ts`); a malformed override could be rejected by some providers. Could validate the minimal `{ effort: low|medium|high|xhigh }` shape and document arbitrary objects as advanced/OpenAI-compatible-only.
- **Field validation**: `inputLimit`/`pricing` should be finite, non-negative numbers.
- **OpenRouter interaction**: decide ordering — explicit user overrides probably apply last, after catalog enrichment.

Happy to send a PR along these lines if the design/naming (`modelOverrides`, user-config-only, known-key overrides) sounds right. Wanted to confirm the shape before implementing.

Environment: oracle 0.15.0 (npm `@steipete/oracle`), API engine, custom OpenAI-compatible `--base-url`.

wangwllu:
## Summary

Closes #273.

Adds a user-config-only `modelOverrides` map so custom OpenAI-compatible gateways (for example LiteLLM-style proxies) can remap a known model's on-wire `apiModel` and model metadata without editing bundled `MODEL_CONFIGS`.

Example:

```json5
{
  "apiBaseUrl": "https://gateway.example/v1",
  "model": "gpt-5.5",
  "modelOverrides": {
    "gpt-5.5": {
      "apiModel": "gateway-model",
      "reasoning": { "effort": "xhigh" },
      "inputLimit": 1050000,
      "pricing": { "inputPerToken": 0.000005, "outputPerToken": 0.00003 }
    }
  }
}
```

## Design

- User config only: project configs cannot set `modelOverrides`, so an untrusted repo cannot reroute model traffic.
- Known models only: tokenizer and unspecified fields are inherited from the existing `MODEL_CONFIGS` entry.
- Narrow field whitelist: `apiModel`, `reasoning`, `inputLimit`, `pricing`.
- Defensive validation for JSON5 input:
  - `apiModel` must be non-empty (trimmed before use).
  - `reasoning: null` explicitly clears a known model's effort; invalid effort is ignored.
  - `inputLimit` must be a positive safe integer.
  - `pricing` must contain finite, non-negative token prices; `pricing: null` clears pricing.
- Applies overrides last, after known/OpenRouter/generic resolution.
- The override `apiModel` also feeds `effectiveModelId` across API CLI, direct/programmatic, and multi-model paths. Browser runs ignore API model overrides.

## Reasoning transport fix

This also fixes a related gap surfaced during review: the OpenAI-compatible chat/completions adapter (used for **all** custom/proxy base URLs — OpenRouter, LiteLLM, vLLM, etc.) only forwarded `model`, `messages`, and `max_tokens`. Any reasoning model's effort was silently dropped on a custom `--base-url`. The adapter now forwards `reasoning_effort` when present.

## Real-gateway behavior proof

Captured the actual on-wire request oracle sends, by pointing it at a local capture server (localhost, dummy key — no secrets). Config under test remaps the known model `gpt-5.5` to a renamed gateway id and pins reasoning effort.

`modelOverrides: { "gpt-5.5": { apiModel: "gateway-model", reasoning: { effort: "xhigh" } } }`

```
CAPTURED_REQUEST {"path":"/v1/chat/completions","model":"gateway-model","reasoning_effort":"xhigh","stream":true}
```

The on-wire `model` is the overridden id (not `gpt-5.5`) and `reasoning_effort` is forwarded.

With `reasoning: null` (clear effort) the same override still remaps the model id but drops the effort:

```
CAPTURED_REQUEST {"path":"/v1/chat/completions","model":"gateway-model","reasoning_effort":null,"stream":true}
```

A live run against a real OpenAI-compatible gateway shows the resolved id in the header (key masked by oracle):

```
Calling gpt-5.5 (API: gateway-model) — 4.26k tokens, no files.
Provider: OpenAI-compatible | base: <redacted-gateway>/v1 | key: OPENAI_API_KEY
Resolved model: gpt-5.5 → gateway-model
Answer:
PROOF-OVERRIDE-OK
```

I also confirmed out-of-band that the gateway honors `reasoning_effort` (reasoning token usage scaled from ~489 at `low` to ~4606 at `xhigh` on the same prompt).

## Tests

- `pnpm run lint`
- `pnpm run build`
- `pnpm test` — full suite green (`1359 passed`, `43 skipped`).
- Maintainer exact-wire smoke: built CLI against an isolated user config and localhost OpenAI-compatible capture server; observed `/v1/chat/completions`, `model: gateway-model`, `reasoning_effort: xhigh`, and `stream: true`, then received `PROOF-OVERRIDE-OK` through the CLI.
- `pnpm run check`, `pnpm run build`, `pnpm run docs:check`, and `pnpm run docs:site`.
- Autoreview: no actionable findings at exact head.
- New coverage: resolver override matrix (apiModel/reasoning/inputLimit/pricing validation, `reasoning: null` clearing, unknown-model no-op), config parse + project-config security boundary + user-override merge survival, runOptions effectiveModelId-from-override (incl. Gemini), and the chat/completions adapter forwarding `reasoning_effort`.

I strip local base-url env vars for the full test run because otherwise provider-routing tests pick up the developer environment's custom endpoints.

## Review notes

This intentionally does not add CLI flags like `--api-model` because those become ambiguous with `--models a,b`. The per-model map keeps multi-model behavior explicit and local to each known model key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Config: add user-only `modelOverrides` in `~/.oracle/config.json` to override a known model's on-wire `apiModel`, reasoning effort, input limit, and pricing — for custom OpenAI-compatible `--base-url` gateways (e.g. a LiteLLM proxy) that rename model ids or need a non-default effort. Overrides apply to known models only (tokenizer inherited), support `reasoning: null` to clear effort, and are ignored from project configs so they cannot reroute model traffic. Fixes #273.
+- API: forward reasoning effort as `reasoning_effort` through the OpenAI-compatible chat/completions adapter used for custom/proxy base URLs (OpenRouter, LiteLLM, vLLM, etc.). Previously the adapter only sent `model`, `messages`, and `max_tokens`, so a reasoning model's effort was silently dropped on any custom `--base-url`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.15.1 — Unreleased
 
+### Added
+
+- Config: add user-only `modelOverrides` in `~/.oracle/config.json` to override a known model's on-wire `apiModel`, reasoning effort, input limit, and pricing — for custom OpenAI-compatible `--base-url` gateways (e.g. a LiteLLM proxy) that rename model ids or need a non-default effort. Overrides apply to known models only (tokenizer inherited), support `reasoning: null` to clear effort, and are ignored from project configs so they cannot reroute model traffic. Fixes #273.
+
 ### Fixed
 
 - Browser: wait up to eight seconds for the ChatGPT model/effort composer pill to mount before failing explicit selection, while leaving `option-not-found` failures immediate. Thanks @gustavosmendes!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@
 
 ### Added
 
-- Config: add user-only `modelOverrides` in `~/.oracle/config.json` to override a known model's on-wire `apiModel`, reasoning effort, input limit, and pricing — for custom OpenAI-compatible `--base-url` gateways (e.g. a LiteLLM proxy) that rename model ids or need a non-default effort. Overrides apply to known models only (tokenizer inherited), support `reasoning: null` to clear effort, and are ignored from project configs so they cannot reroute model traffic. Fixes #273.
-- API: forward reasoning effort as `reasoning_effort` through the OpenAI-compatible chat/completions adapter used for custom/proxy base URLs (OpenRouter, LiteLLM, vLLM, etc.). Previously the adapter only sent `model`, `messages`, and `max_tokens`, so a reasoning model's effort was silently dropped on any custom `--base-url`.
+- API: add user-only `modelOverrides` for remapping known models and their metadata on custom OpenAI-compatible gateways. Fixes #273. Thanks @wangwllu!
 
 ### Fixed
 
+- API: forward configured reasoning effort through custom OpenAI-compatible chat-completions gateways.
 - Browser: wait up to eight seconds for the ChatGPT model/effort composer pill to mount before failing explicit selection, while leaving `option-not-found` failures immediate. Thanks @gustavosmendes!
 
 ## 0.15.0 — 2026-06-19
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -17,10 +17,11 @@ import chalk from "chalk";
 import type { SessionMetadata, SessionMode, BrowserSessionConfig } from "../src/sessionStore.js";
 import { sessionStore, pruneOldSessions } from "../src/sessionStore.js";
 import { DEFAULT_MODEL, MODEL_CONFIGS } from "../src/oracle/config.js";
-import { isKnownModel } from "../src/oracle/modelResolver.js";
+import { isKnownModel, resolveOverriddenApiModel } from "../src/oracle/modelResolver.js";
 import type {
   ApiProviderMode,
   ModelName,
+  ModelOverridesConfig,
   PreviewMode,
   RunOracleOptions,
 } from "../src/oracle/types.js";
@@ -195,6 +196,7 @@ type ResolvedCliOptions = Omit<CliOptions, "model"> & {
   model: ModelName;
   models?: ModelName[];
   effectiveModelId?: string;
+  modelOverrides?: ModelOverridesConfig;
   writeOutputPath?: string;
   previousResponseId?: string;
   followupSessionId?: string;
@@ -1328,6 +1330,7 @@ function buildRunOptions(
     browserResumeConversationUrl:
       overrides.browserResumeConversationUrl ?? options.browserResumeConversationUrl,
     effectiveModelId: overrides.effectiveModelId ?? options.effectiveModelId ?? options.model,
+    modelOverrides: overrides.modelOverrides ?? options.modelOverrides,
     file: overrides.file ?? options.file ?? [],
     maxFileSizeBytes: overrides.maxFileSizeBytes ?? options.maxFileSizeBytes,
     slug: overrides.slug ?? options.slug,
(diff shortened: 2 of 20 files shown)


wangwllu:
@clawsweeper re-review

Addressed both P1 asks from the prior review:

1. **Reasoning transport fixed** (was the core defect): the OpenAI-compatible chat/completions adapter (`buildOpenRouterCompletionClient`) now forwards `reasoning_effort`. Previously it only sent `model`/`messages`/`max_tokens`, dropping effort on every custom `--base-url`. Added adapter unit tests for forward + omit cases.

2. **Real behavior proof added** to the PR body: captured the actual on-wire request (localhost capture server, no secrets) showing the overridden `model` id (`GPT-5.5-joybuilder`, not `gpt-5.5`) and `reasoning_effort: "xhigh"` on `/v1/chat/completions`, plus the `reasoning: null` clear case and a redacted live real-gateway run header (`Resolved model: gpt-5.5 → GPT-5.5-joybuilder`).

Also fixed a missed CLI caller: `bin/oracle-cli.ts` builds run options independently of `resolveRunOptionsFromConfig`, so it now threads `modelOverrides` and resolves the override `apiModel` into `effectiveModelId` (the real-gateway run surfaced this).

lint + build clean; full suite green in isolation.

steipete:
Maintainer verification for exact head `e6d5d509f6aa75c57e5c7279ae9591128d7402b5`:

- Accepted and fixed one autoreview finding: detached/restarted API sessions now persist and rehydrate `modelOverrides`, so the default background Pro path matches `--wait` behavior.
- Added an integration regression that deletes the user config before `--exec-session` and verifies the stored on-wire model override and reasoning effort are still applied.
- Focused proof: `pnpm vitest run tests/oracle/modelResolver.override.test.ts tests/oracle/clientFactory.test.ts tests/config.test.ts tests/runOptions.test.ts` (80 passed); `pnpm vitest run tests/sessionManager.test.ts` (20 passed); focused detached-session integration test (passed).
- Repository gates: `pnpm run check`, `pnpm run build`, and `pnpm run docs:check` passed locally.
- Fresh autoreview: clean, no accepted/actionable findings.
- Exact-head hosted CI: Ubuntu, macOS, Windows, and GitGuardian passed: https://github.com/steipete/oracle/actions/runs/28522037563
- Public model-identifier audit: PASS; the diff uses documented repository model names and generic gateway placeholders only.

The user-config-only boundary, known-model restriction, validated field whitelist, and redacted gateway proof make the new routing contract supported for landing.

steipete:
Landed as `ea453c7b9483cefc23e38d533a1c7a24b4741886`.

Final proof: 80 focused resolver/config/client/run-option tests, 20 session-manager tests, detached `--exec-session` regression, `pnpm run check`, `pnpm run build`, `pnpm run docs:check`, clean autoreview, and exact-head Ubuntu/macOS/Windows/GitGuardian CI: https://github.com/steipete/oracle/actions/runs/28522037563

The shipped shape keeps routing overrides user-config-only, limited to known models and validated fields, and now preserves them across detached/restarted sessions. Thanks @wangwllu!

PR #278, opened 2026-06-29
fix(browser): avoid truncating Pro/Pro-Extended answers captured mid-stream at the thinking→answer transition

xuan-wei:
## Summary

- Prevent browser-mode Pro answers from being returned during ChatGPT's thinking-to-answer transition, when completion controls can appear before response text is stable.
- Route every completion-control capture through the existing stability watchdog instead of trusting a fixed text-length threshold; partial first paragraphs can exceed any small threshold.
- Never replace the already captured candidate with a shorter watchdog result.
- Cover a partial capture longer than 80 characters growing into the complete answer, plus a legitimate short answer remaining intact after stability confirmation.

## Verification

- `pnpm vitest run tests/browser/pageActions.test.ts` — 58 passed, 1 intentional skip.
- `pnpm run check` — passed.
- `pnpm run build` — passed.
- Autoreview — clean after replacing the original sub-80-character heuristic with unconditional completion-control stability confirmation.

## Live ChatGPT proof

Built this exact implementation and ran authenticated Pro browser captures:

- Long-response smoke: Oracle waited through Pro thinking, captured all eight requested numbered items, and included the unique required final-tail marker `ORACLE_PR278_FINAL_TAIL_OK_20260702` (261 output tokens, 55.2 seconds).
- Short-response smoke: Oracle returned the exact requested marker `ORACLE_PR278_SHORT_OK_20260702` intact.

The external transition race is timing-dependent; deterministic regression tests exercise the completion-control path with both a >80-character partial capture and a stable short answer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - API: forward configured reasoning effort through custom OpenAI-compatible chat-completions gateways.
 - Browser: accept a stable, exact file-input name match when ChatGPT marks the composer ready but exposes no attachment chip or count, while still waiting through active uploads and rejecting missing or extra files. Fixes #275. Thanks @wangwllu!
+- Browser: avoid returning truncated Pro answers when completion controls appear during the thinking-to-answer transition. Thanks @xuan-wei!
 - Browser: wait up to eight seconds for the ChatGPT model/effort composer pill to mount before failing explicit selection, while leaving `option-not-found` failures immediate. Thanks @gustavosmendes!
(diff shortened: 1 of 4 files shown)


steipete (approved):
Maintainer review complete. I strengthened the transition guard to confirm stability for every completion-control capture, because partial text can exceed a fixed threshold. Focused tests, full checks/build, clean autoreview, and authenticated Pro long-tail plus short-answer browser smokes passed on the updated head.

Issue #286, opened 2026-07-03
Browser mode: structured no-file prompts can fail commit verification before promptSubmitted


derekszen:
## Summary

On current `main`, ChatGPT browser mode can fail during prompt send/commit verification for a moderately structured prompt with **no file attachments**, even though tiny/plain prompts in the same profile and model work immediately.

The failing run reports:

```text
ERROR: Prompt did not appear in conversation before timeout (send may have failed)
User error (browser-automation): Prompt did not appear in conversation before timeout (send may have failed)
```

The stored session metadata for the failed run has only `browser.config`; it does **not** persist `browser.runtime`, `conversationId`, `tabUrl`, `promptSubmitted`, or model-selection evidence. Successful runs in the same setup do persist `browser.runtime.promptSubmitted=true` and a ChatGPT `/c/<id>` URL.

This looks distinct from:

- #10: closed historical prompt-commit false negative for long prompts/attachments.
- #115 / #116: closed attachment/send-readiness race; this repro uses no files.
- #284 / #285: after-submit thinking/liveness/harvest issues; this repro fails before Oracle confirms submission.
- #283: stale `conv*` cookie cleanup; I built and tested that PR locally and the same compact prompt still failed.

## Environment

- Oracle version: `0.15.0`
- Current main tested: `bda0326d43b02c5346e742692865fc21d8c5fc35` (`Add secure bridge artifact transfer (#277)`)
- Also tested PR #283 build: `a02eb536aa2ed1f707e91514013184bba8b23876`; same failure
- OS: Arch Linux, kernel `7.0.13-arch1-1`, x86_64
- Browser: `Chromium 149.0.7827.200 Arch Linux`
- Node: `v25.9.0`
- pnpm: `10.33.2`
- Mode: local browser foreground, manual-login persistent profile
- Model: `gpt-5.5-pro`, explicit selector strategy (`modelStrategy=select`), desired label `Pro`

## What succeeds

Tiny exact-response smoke succeeds and verifies model selection:

```bash
oracle --engine browser --model gpt-5.5-pro \
  -p 'Reply with exactly: ORACLE_GPT55_SELECT_OK'
```

Observed metadata:

```json
{
  "status": "completed",
  "browser.runtime.promptSubmitted": true,
  "browser.runtime.tabUrl": "https://chatgpt.com/c/...",
  "browser.modelSelection": {
    "requestedModel": "Pro",
    "resolvedLabel": "Pro",
    "strategy": "select",
[...]

PR #291, opened 2026-07-04
fix(browser): keep ChatGPT turn indexes consistent

cp7553479:
## Problem

Current ChatGPT markup nests role-bearing message nodes inside top-level `conversation-turn` containers. The legacy broad selector matches both layers, so browser baselines and later response/artifact readers can count different turn indexes. Image-only replies are especially vulnerable: the new response is rendered, but the `minTurnIndex` gate can suppress it until timeout.

## Fix

- Add one shared precise-first conversation-turn expression, retaining the broad selector only as a legacy-markup fallback.
- Use that same coordinate system for prompt commit verification, response snapshots and markdown fallback, image/file artifact capture, attachment checks, Deep Research, and reattachment.
- Resolve fallback content and iframe owners by containment in the selected top-level turn list, avoiding nested-node index drift.
- Add focused regressions for nested broad-selector matches, reused conversations, assistant turn indexes, and legacy fallback behavior.
- Add an Unreleased changelog entry thanking @cp7553479.

## Verification

- `pnpm run check`
- `pnpm run build`
- `pnpm test`: 1,401 passed, 43 skipped
- Focused rebased-head browser suite: 139 passed
- Structured pre-ship review: clean, no actionable findings
- Authenticated live Chrome/CDP test: one initial text turn plus a same-conversation image follow-up completed in 64.7s; Oracle captured the follow-up and saved a real 1,768,098-byte PNG at 1254×1254.

The broad selector remains available for older/project DOM shapes; current ChatGPT markup now uses one top-level index from submission through artifact capture.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Browser: persist late `/c/<id>` URLs during remote Chrome runs and prefer saved conversation targets over stale target IDs during reattach. Fixes #284. Thanks @LeoLin990405 and @StartupBros!
+- Browser: keep prompt baselines, assistant snapshots, and artifact capture on one top-level ChatGPT turn index so nested message nodes cannot hide a new response. Thanks @cp7553479!
 
 ## 0.15.1 — 2026-07-03
 
--- a/src/browser/actions/assistantResponse.ts
+++ b/src/browser/actions/assistantResponse.ts
@@ -7,6 +7,7 @@ import {
   FINISHED_ACTIONS_SELECTOR,
   STOP_BUTTON_SELECTORS,
 } from "../constants.js";
+import { buildConversationTurnListExpression } from "../conversationTurns.js";
 import { delay } from "../utils.js";
 import {
   logDomFailure,
@@ -607,8 +608,7 @@ async function isCompletionVisible(Runtime: ChromeClient["Runtime"]): Promise<bo
           return Boolean(node.querySelector(ASSISTANT_SELECTOR) || node.querySelector('[data-testid*="assistant"]'));
         };
 
-        const preciseTurns = Array.from(document.querySelectorAll('[data-testid^="conversation-turn"]'));
-        const turns = preciseTurns.length > 0 ? preciseTurns : Array.from(document.querySelectorAll('${CONVERSATION_TURN_SELECTOR}'));
+        const turns = ${buildConversationTurnListExpression()};
         let lastAssistantTurn = null;
         for (let i = turns.length - 1; i >= 0; i--) {
           if (isAssistantTurn(turns[i])) {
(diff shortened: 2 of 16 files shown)


LeoLin990405 (commented):
Solid root-cause analysis, and I can confirm the direction is right — the double-count trap is real and the precise-preferred/broad-fallback pattern is the correct shape. Verified locally on this branch: `pnpm run typecheck` clean, `assistantResponseStatus` + `pageActions` all green, full suite 1389 passed / 0 failed.

Two consumers of the same invariant look like they still need the treatment, though. The PR's own thesis is "baseline count and turn-index consumers must read the same list" — but two gated consumers still read the broad list against the now-precise baseline:

**1. The root-scored markdown fallback (`assistantResponse.ts`, `resolveTurnIndex` / `isAfterMinTurn`)**

```js
const CONVERSATION_SELECTOR = '${CONVERSATION_TURN_SELECTOR}';
const turnNodes = Array.from(document.querySelectorAll(CONVERSATION_SELECTOR));
...
const isAfterMinTurn = (node) => { ... idx >= __minTurn; };
```

This gate now compares **broad indexes** against a **precise baseline**, which inverts the original bug: on a DOM where each `section[data-testid^="conversation-turn"]` wraps a nested role div, broad indexes run ~2× the precise ones, so a *pre-existing* assistant turn (broad index 3, precise baseline 2) passes `isAfterMinTurn` and the fallback can capture a stale answer as the new one. Before this PR the inflation was at least consistent on both sides. Same precise-preferred pattern on `turnNodes` fixes it.

**2. `verifyPromptCommitted` (`promptComposer.ts`)**

Its in-page script counts turns with the broad selector, but its `baselineTurns` comes from `readConversationTurnCount` — precise after this PR. Result: `hasNewTurn` can be spuriously true before the send even lands (broad count 3 > precise baseline 2 with zero new turns), which weakens both the `matchesPrompt && hasNewTurn` path and the `composerCleared && hasNewTurn && …` fallback commit — the safety net that #286/#288 just made observable. Same fix applies.

Two smaller notes:

- The precise selector is hardcoded as a string literal in the three `assistantResponse.ts` sites while `constants.ts` exports `CONVERSATION_TURN_CONTAINER_SELECTOR` (only `index.ts` imports it). Using the constant everywhere keeps the pattern greppable for the next UI-drift fix.
- "the 3 pre-existing `pageActions` model-selector failures … present on `main`" — I can't reproduce that: on current `main` (`c159bf8`, Node 26, macOS) `pageActions.test.ts` is 58 passed / 0 failed. Possibly a stale local base? Worth re-checking so the claim doesn't confuse triage.

(I attempted an independent live-DOM count of precise vs broad on a real conversation to double-confirm your table, but my network path to chatgpt.com dropped mid-check — your CDP numbers plus the DOM structure make the case regardless.)

Issue #284, opened 2026-07-03
Browser mode: thinking-state detection can miss (ChatGPT UI drift) — live runs look dead; reattach then binds a stale tab target


StartupBros:
TITLE: Browser mode: thinking-state detection can miss (ChatGPT UI drift) — live runs get misclassified as stalled; harvest then binds a stale tab target

Seen on 0.15.0, browser mode against a remote CDP Chrome (Xvfb, Linux/WSL2), model gpt-5.5-pro (Pro Extended), 2026-07-02.

**What happened**
1. Prompt submitted fine; ChatGPT created the conversation and began a long thought (11m14s, title "GitHub PR Deep Review").
2. oracle logged `Waiting for ChatGPT response - Xm elapsed; no thinking status detected yet` for the entire run — the thinking indicator was never recognized despite the model actively thinking.
3. Our wrapper killed the CLI on a no-think watchdog (assuming a dead submission). `oracle session <slug>` reattach then reported `Assistant turns: 0` with `URL: https://chatgpt.com/` — the stored tab target no longer pointed at the conversation tab, so harvest returned nothing even after the response completed.
4. The finished response was fully present in the conversation tab's DOM; we recovered it by reading `document.body.innerText` over CDP and matching the output grammar.

**Suggested hardening**
- Fall back to outcome-based detection: if the thinking indicator is not seen within N seconds but a conversation URL (`/c/<id>`) exists for the session, poll the DOM for assistant-turn content instead of relying on the indicator selector.
- On reattach/harvest, re-resolve the tab by conversation URL (stored at submit time) rather than trusting the recorded CDP target id, which goes stale across kills/restarts.

Happy to share the full run log privately. Our downstream workaround (DOM polling for the output grammar + a probe that distinguishes "tab exists = quota spent, never resubmit" from "truly dead") is in our wrapper if useful as a reference.

StartupBros:
Additional evidence, different failure shape (same run environment, 0.15.0, remote-CDP browser mode, gpt-5.5-pro):

- Thinking status flapped during one review: `no thinking status detected` (0-1m) -> `status=active; source=inline` (1m30-2m30) -> back to `no thinking status detected` (3m-6m) while generation continued.
- At ~6m the CLI concluded the answer was complete, captured a single character (`Answer: I`, download counter `↓1`), archived the conversation, and reported success (`8m46s · gpt-5.5-pro[browser] · ↑1.29k ↓1`).
[...]

PR #293, opened 2026-07-06
fix(browser): require completion proof for short captures

LeoLin990405:
Addresses the remaining premature-capture path from #284.

## Failure

ChatGPT can briefly expose a one-character assistant turn while replacing its thinking UI. During that window, stop and completion controls can both be absent. Stability alone is not completion proof: the first live injection reproduced the race and showed that a stable `I` could still be finalized and archived.

## Fix

- Treat responses under 16 characters as suspicious across observer, watchdog, and recovery paths.
- While a response remains suspiciously short, require visible completion controls; ordinary text stability is insufficient.
- If the response grows beyond the suspicious threshold, retain the existing stability behavior.
- Fail closed when a short response cannot be confirmed, and bound a hung observer when its watchdog expires.
- Keep the threshold shared between TypeScript polling and the generated browser observer.

## Verification

- `pnpm run check`
- `pnpm test`: 1,410 passed, 43 skipped
- `pnpm run build`
- Autoreview clean after one accepted return-path finding and regression fix

Authenticated Chrome/CDP proof used a reversible injector to hide stop/completion controls and force the newest assistant DOM to exactly `I` for six seconds. Oracle rejected that stub through 76 forced cycles; after release, the DOM recovered to 921 characters, Oracle returned the full answer with marker `ORACLE293-FINAL-7C91`, saved the transcript, and only then archived the conversation. Injector and Oracle both exited successfully.

Thanks @LeoLin990405 for the original diagnosis and patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Browser: persist late `/c/<id>` URLs during remote Chrome runs and prefer saved conversation targets over stale target IDs during reattach. Fixes #284. Thanks @LeoLin990405 and @StartupBros!
 - Browser: keep prompt baselines, assistant snapshots, and artifact capture on one top-level ChatGPT turn index so nested message nodes cannot hide a new response. Thanks @cp7553479!
+- Browser: reconfirm implausibly short ChatGPT captures after thinking-UI transitions, and fail closed rather than archiving when the response cannot be confirmed. Fixes #284. Thanks @LeoLin990405!
 
 ## 0.15.1 — 2026-07-03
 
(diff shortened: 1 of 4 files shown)


LeoLin990405:
Added a live-proof section to the PR body: a real signed-in `gpt-5.5-pro[browser]` run with a sub-16-char answer shows the guard path firing (`Captured an implausibly short response; confirming…`), the confirmation running before transcript-save and archive, and the final `OK` answer preserved (no regression on legitimate short replies). @clawsweeper re-review

PR #299, opened 2026-07-08
feat(browser): ORACLE_BROWSER_MAX_CONCURRENT_TABS env override for the tab-concurrency cap

StartupBros:
## Gap

`maxConcurrentTabs` is already configurable via the config file (`browser.maxConcurrentTabs`) and the `--browser-max-concurrent-tabs` flag — but unlike its sibling browser knobs it has no environment override. In `resolveBrowserConfig`, `ORACLE_BROWSER_PORT`/`ORACLE_BROWSER_DEBUG_PORT`, `ORACLE_BROWSER_ALLOW_COOKIE_ERRORS`, and `ORACLE_BROWSER_PROFILE_DIR` all consult `process.env`; the tab cap falls straight through to the default.

## Change

- `ORACLE_BROWSER_MAX_CONCURRENT_TABS` env fallback with the same precedence as the siblings: **explicit config > env > default**.
- Invalid / non-positive values are ignored (`parseMaxConcurrentTabs`, mirroring `parseDebugPort`; `normalizeMaxConcurrentTabs` still guards the final value).
- Test in `tests/browser/config.test.ts` following the existing `ORACLE_BROWSER_PROFILE_DIR` save/restore pattern (config-beats-env, env-beats-default, `0`/garbage ignored).
- One-line docs note in `docs/configuration.md`, same style as the `manualLoginProfileDir` env hint.

## Why

Orchestrators that drive oracle across hosts (we run it as a PR-review gate, per-host concurrency tuned by an adaptive governor) set per-host env rather than editing `~/.oracle/config.json` — this makes the tab cap tunable the same way the other browser knobs already are.

`vitest run tests/browser/config.test.ts`: 6/6 pass; `tsc --noEmit` clean.

(Follow-up in the same area as #285, which you kindly merged — thanks!)

## Real behavior proof (added per review)

Patched build, real signed-in remote-Chrome session (`--remote-chrome 127.0.0.1:9222`), manual-login profile, user config's `browser.maxConcurrentTabs` key removed so nothing shadows the env var. Same invocation, only the env var differs:

**Baseline — no env var (built-in default):**
```
$ node dist/bin/oracle-cli.js --engine browser --remote-chrome 127.0.0.1:9222 \
    --browser-model-strategy current --no-notify -p "Reply with exactly: proof-charlie"
[browser] Acquired ChatGPT browser slot d190a21b (3 max).
```

**With the override:**
```
$ ORACLE_BROWSER_MAX_CONCURRENT_TABS=5 node dist/bin/oracle-cli.js --engine browser \
    --remote-chrome 127.0.0.1:9222 --browser-model-strategy current --no-notify \
    -p "Reply with exactly: proof-delta"
[browser] Acquired ChatGPT browser slot 0ff83ccc (5 max).
```

(Full logs available; both proof runs later hit a ChatGPT-side "prompt did not appear" send flake unrelated to config resolution — the tab-lease acquisition above happens from the resolved config exactly as in clean runs. Happy to re-capture a full clean transcript on request. Precedence also confirmed live: with `browser.maxConcurrentTabs: 4` present in the user config, the same env-var run resolves `(4 max)` — config beats env, as designed.)

## Review follow-ups addressed

- **Test isolation** (ClawSweeper P3 / Codex P2): `beforeEach` now clears `ORACLE_BROWSER_MAX_CONCURRENT_TABS` so a developer/CI export can't leak into the default assertions; `afterEach` still restores the caller's original value. Suite verified green with and without a hostile pre-set value (`ORACLE_BROWSER_MAX_CONCURRENT_TABS=7 vitest run …` → 6/6).


StartupBros:
Both review asks addressed in `161b58b` + the updated PR body:

1. **Test isolation** (Codex P2 / ClawSweeper P3): `beforeEach` clears `ORACLE_BROWSER_MAX_CONCURRENT_TABS` before every test so a caller's export can't reach the default assertions; `afterEach` still restores the original value. Verified green with a hostile `=7` pre-set.
2. **Real behavior proof** (ClawSweeper P1): added to the PR body — same patched build, same live remote-Chrome invocation, `(3 max)` without the env var vs `(5 max)` with it, plus live confirmation that an explicit `browser.maxConcurrentTabs` config key still beats the env var.

Thanks for the fast automated review!

StartupBros:
@clawsweeper re-review

PR #300, opened 2026-07-09
fix(cli): recover Deep Research App tool-call wrapper captures via reattach

### LINKED ISSUE #297: Placeholder-capture detection misses 'Deep Research App' wrapper, so session --render replays junk instead of live-reattaching
## Summary

On oracle 0.15.2 (macOS, browser mode, manual-login profile), a
`--browser-research deep` run against ChatGPT's new "Deep Research App" UI
saved wrapper text instead of the report, and `oracle session <id> --render`
replayed the bad capture rather than triggering the live OOPIF reattach that
would have recovered it.

## Repro

1. `oracle --engine browser --browser-manual-login --browser-research deep
   --write-output out.md -p "<research prompt>"` (oracle 0.15.2,
   gpt-5.5-pro, Pro plan).
2. ChatGPT runs the research through the "Deep Research App" connector; the
   report renders inside `connector_openai_deep_research.web-sandbox.
   oaiusercontent.com` (nested `iframe#root`). The conversation DOM shows a
   tool-call wrapper: "Called tool / Deep Research App / Call tool / Request
   ... Response { session_id: ... }".
3. The run ends after ~19m; `deep-research-report.md` and `--write-output`
   both contain the wrapper text (~23 lines), not the report. The report was
   visibly complete in the tab ("Research completed in 18m").
4. `oracle session <id> --render` prints the same wrapper junk; no reattach.

## Suspected cause

`isDeepResearchPlaceholderCapture` (src/cli/sessionDisplay.ts:66) treats a
capture as recoverable only when the trimmed answer exactly equals
"called tool"/"used tool" (plus variants) and outputTokens <= 8. The
"Called tool / Deep Research App" wrapper (with request/response JSON) fails
that exact-equality check, so `canReattach` never fires and the stale
artifact is replayed. A substring/structure heuristic (e.g. any capture
whose text contains "Called tool" within the first line, or whose text lacks
report structure while the session used deep research) would catch this
variant.

## Confirmation the OOPIF path works

Attaching over CDP to the sandbox iframe target and reading
`iframe#root.contentDocument` recovered the full report (33,918 chars) with
its evidence table intact — i.e. the 0.15.2 target-attach capture
architecture is correct; only the recovery trigger is too narrow.

## Environment

macOS 25.5.0, Chrome stable, oracle 0.15.2 (npm), manual-login profile,
ChatGPT Pro, model gpt-5.5-pro, Deep Research via composer tools menu.

devYRPauli:
Fixes #297.

## Problem

The placeholder detector recognized only a bare `Called tool` answer with at most eight output tokens. ChatGPT's Deep Research App connector can instead save a multi-line tool-call wrapper containing `Deep Research App`, request/response fields, and a session id. That wrapper was replayed as the answer instead of triggering the existing reattach path that can recover the completed report.

The proposed prefix-only match was too broad: a legitimate report beginning with prose such as “Called tool adoption…” could be treated as incomplete and converted from completed to error. An already-recovered wrapper also remained ahead of the recovered report during rendering.

## Final design

- Preserve the existing token gate for bare localized tool-call stubs.
- Recognize a multi-line placeholder only when the first line is an exact known marker and the remaining lines contain the actual Deep Research App wrapper structure.
- Do not classify prose that merely begins with a marker phrase.
- Do not reattach again after a recorded reattach answer.
- When rendering an already-recovered wrapper session, skip the stale wrapper and select the later recovered `Answer:` block.

Contributor commits and authorship are preserved. The maintainer hardening commit adds the false-positive guard, recovered-render behavior, tests, and changelog entry.

## Validation

Exact final head: `ccd582bc326a0486cc525fbf83859cb4cae70924`.

- `pnpm vitest run tests/cli/sessionDisplay.test.ts tests/browser/pageActions.test.ts` — 107 passed, 1 skipped.
- `pnpm check` — passed.
- `pnpm test` — 1,502 passed, 43 skipped.
- Final AutoReview — no actionable findings.

## Exact-head live Pro proof

Signed-in ChatGPT Pro Deep Research session `pr300-deep-research-live` completed a sourced report in 2m07s and included the unique final marker `PR300-DEEP-RESEARCH-END-20260711`.

`oracle session pr300-deep-research-live --render --verbose-render` then:

- kept the session `completed`;
- rendered the complete report through the unique final marker;
- validated the saved Deep Research report and browser transcript artifacts;
- did not misclassify the normal report or convert it to an error.

Fresh exact-head CI is running.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Browser: keep hidden macOS Chrome windows rendered off-screen so trusted prompt submissions land without retaining drafts or leaking them into later runs. Fixes #298 and #312. Thanks @LeoLin990405!
 - Browser: require positive terminal evidence before finalizing ChatGPT responses so settled preambles and mid-stream text cannot be captured as the completed answer. Thanks @StartupBros!
 - Browser: distinguish genuine Cloudflare interstitials from healthy ChatGPT pages that carry bot-management scripts or mention generic challenge text. Thanks @StartupBros!
+- Browser: recover completed Deep Research reports when the initial capture contains only the Deep Research App tool-call wrapper. Thanks @devYRPauli!
 
 ## 0.15.2 — 2026-07-06
 
--- a/src/cli/sessionDisplay.ts
+++ b/src/cli/sessionDisplay.ts
@@ -70,32 +70,40 @@ const DEEP_RESEARCH_TOOL_CALL_MARKERS = [
   "narzędzie wywołane",
 ];
 
+function isDeepResearchToolCallPlaceholder(answerText: string, outputTokens?: number): boolean {
+  const lines = answerText
+    .toLowerCase()
+    .split(/\r?\n/)
+    .map((line) => line.replace(/\s+/g, " ").trim())
+    .filter(Boolean);
+  if (!lines[0] || !DEEP_RESEARCH_TOOL_CALL_MARKERS.includes(lines[0])) {
+    return false;
+  }
+  if (lines.length === 1) {
+    return outputTokens == null || outputTokens <= 8;
+  }
+  const wrapper = lines.slice(1).join(" ");
+  const structuralSignals = [
+    wrapper.includes("deep research app"),
+    /\bcall tool\b/.test(wrapper),
+    /\brequest\s*\{/.test(wrapper),
+    /\bresponse\s*\{/.test(wrapper),
+    /\bsession[_ ]id\b/.test(wrapper),
+  ].filter(Boolean).length;
+  return wrapper.includes("deep research app") && structuralSignals >= 2;
+}
+
 export function isDeepResearchPlaceholderCapture(
   metadata: SessionMetadata,
   logText: string,
 ): boolean {
-  const answer = trimBeforeFirstAnswer(logText)
-    .replace(/^Answer:\s*/i, "")
-    .toLowerCase()
-    .replace(/\s+/g, " ")
-    .trim();
-  // A capture whose answer begins with a tool-call marker is a placeholder, not the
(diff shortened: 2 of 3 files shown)


devYRPauli:
Good catch on the recovered-log regression. Fixed in f7d9d8d5: `isDeepResearchPlaceholderCapture` now returns false when a later `Answer:` already exists in the log, so an already-recovered multi-line wrapper (where `trimBeforeFirstAnswer`'s bare-stub skip does not apply) no longer re-triggers reattach over a completed session. Added a regression test for that shape (wrapper + `[reattach]` + recovered `Answer:` -> not a placeholder); it fails on the previous commit and passes now. Full sessionDisplay suite is 34/34, `tsc --noEmit` and `oxlint` clean.

On real-behavior proof: I do not have a ChatGPT Pro Deep Research setup to capture a live redacted `oracle session <id> --render` transcript, so I cannot post one here. The unit tests cover the exact captured-log shapes (fresh wrapper triggers reattach; recovered wrapper does not).

steipete:
Maintainer hardening and live proof complete at `ccd582bc326a0486cc525fbf83859cb4cae70924`.

- Wrapper detection now requires the actual Deep Research App structure; ordinary prose beginning “Called tool…” stays completed.
- Already-recovered sessions render the recovered report rather than the stale wrapper.
- Focused: 107 passed, 1 skipped.
- `pnpm check`: passed.
- Full suite: 1,502 passed, 43 skipped.
- Final AutoReview: no actionable findings.
- Signed-in Pro Deep Research: completed in 2m07s; `oracle session pr300-deep-research-live --render --verbose-render` rendered the complete report and validated both saved artifacts without changing the completed status.

Fresh exact-head CI is running now.

PR #307, opened 2026-07-10
Add Requesty as an OpenAI-compatible provider

Thibaultjaigu:
## What

Adds Requesty as an OpenAI-compatible provider. [Requesty](https://requesty.ai) is an LLM gateway that exposes an OpenAI-compatible API and uses the same `provider/model` id convention as OpenRouter, so this mirrors the existing OpenRouter wiring as closely as possible.

## Changes

- **`src/oracle/modelResolver.ts`** — detect/normalize the Requesty base URL, add a default base (`https://router.requesty.ai/v1`), and hydrate model metadata from the Requesty `/v1/models` catalog. Requesty reports context length as `context_window` and prices as flat per-token USD, which is mapped to the internal model-info shape.
- **`src/oracle/client.ts`** — route Requesty base URLs through the OpenAI chat/completions adapter (like the other proxy providers) and forward optional attribution headers `HTTP-Referer` / `X-Title` from `REQUESTY_REFERER` / `REQUESTY_TITLE`, matching the OpenRouter path.
- **`src/oracle/providerRoutePlan.ts`, `src/oracle/run.ts`, `src/cli/sessionRunner.ts`** — add a Requesty gateway fallback and key source (`REQUESTY_API_KEY`). OpenRouter keeps precedence when both keys are set, so existing setups are unchanged; Requesty takes over the automatic gateway fallback only when `OPENROUTER_API_KEY` is absent but `REQUESTY_API_KEY` is present.
- **`docs/requesty.md`** + a Requesty row in the `docs/install.md` provider table.
- **`tests/oracle/providerRoutePlan.test.ts`** — cover the Requesty fallback, the OpenRouter-precedence rule, and explicit Requesty base URL routing.

## Usage

```bash
export REQUESTY_API_KEY="rqsty-sk-..."
oracle --engine api --model openai/gpt-4o-mini -p "..."
# or explicitly:
oracle --base-url https://router.requesty.ai/v1 --model anthropic/claude-sonnet-4-5 -p "..."
```

## Testing

- `pnpm run typecheck` — clean.
- `oxlint` + `oxfmt --check` on the edited files — clean.
- `vitest run` on the affected suites (`providerRoutePlan`, `modelResolver.override`, `providerFailures`, `clientFactory`, `run.render`) — 53 passed, including the 3 new Requesty tests.
- Live-tested against the real endpoint: a chat/completions call to `https://router.requesty.ai/v1` with the attribution headers returned successfully, and `GET /v1/models` returned 605 models with `context_window: 128000` / `supports_tool_calling: true` for `openai/gpt-4o-mini`, confirming the catalog field mapping.

Docs: https://docs.requesty.ai · API keys: https://app.requesty.ai/api-keys · Model list: https://app.requesty.ai/router/list

---

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust or close it if it's not a fit.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -46,6 +46,7 @@ API mode is opt-in and reads keys from the environment. Set whichever providers 
 | Google       | `GEMINI_API_KEY`                                                  | Gemini 3.1 Pro, Gemini 3.5 Flash, Gemini 3.1 Flash-Lite |
 | Anthropic    | `ANTHROPIC_API_KEY`                                               | Claude Sonnet 4.6, Claude Opus 4.1                      |
 | OpenRouter   | `OPENROUTER_API_KEY`                                              | Any OpenRouter id (e.g. `minimax/minimax-m2`)           |
+| Requesty     | `REQUESTY_API_KEY`                                               | Any Requesty id (e.g. `openai/gpt-4o-mini`)             |
 
 If no key is set, Oracle defaults to **browser mode** and drives ChatGPT directly — see [Browser Mode](browser-mode.md) for the manual-login flow.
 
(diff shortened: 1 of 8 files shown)


steipete:
I have never heard of this company. If they become popular, happy to revisit.

PR #309, opened 2026-07-10
fix(browser): scope the aria-label stop fallback to the composer form (follow-up to #285)

StartupBros:
Follow-up to the post-merge review note on #285.

## Problem

The `button[aria-label*="stop" i]` fallback added in #285 matches document-wide, so any visible non-generation stop control (read-aloud, voice mode, dictation) keeps `isStopButtonVisible()` true. `pollAssistantCompletion()` only finalizes when that value goes false, so a completed browser response could stall until the response timeout.

## Fix

Scope the fallback to the composer form and exclude the known non-generation stop controls that legitimately render there: `form button[aria-label*="stop" i]:not([aria-label*="dictat" i]):not([aria-label*="voice" i]):not([aria-label*="read" i])`. The `data-testid` selectors stay primary; the fallback still covers testid drift for the real generation stop, which lives inside the composer form.

## Validation

`pnpm run typecheck`, `oxlint`, `oxfmt --check`, full browser suite (626 passing). The fake DOM in `thinking.test.ts` gained faithful descendant-combinator and `:not()` matching so the scoped selector is actually exercised; new regressions cover read-aloud/dictation/voice stops inside and outside the form.

## Real-behavior proof (redacted)

**Live CDP capture during an ACTIVE generation on a production GPT-5.6 conversation tab (2026-07-10 14:5x EDT):**

```json
{"generating":true,"testidStop":1,"oldSelectorMatches":1,"oldSelectorLabels":["Stop answering"],"newSelectorMatches":1,"newSelectorLabels":["Stop answering"],"pageWideStopLabels":["Stop answering"],"cf":{"hasAppShell":true,"hasChallengeScript":true,"bodyLen":59167,"titleIsChallenge":false},"visibleProgressBars":0}
```

The real generation stop control (`aria-label="Stop answering"`) lives inside the composer form on the live DOM: the scoped fallback matches it 1:1 with the old document-wide selector (`newSelectorMatches: 1`, same node), the `data-testid="stop-button"` primary is present, and form ancestry holds exactly as the selector assumes.

**Behavioral evidence from the fork running this change:** reviews completing after the deploy (09:04, 13:31, 14:14 EDT on 2026-07-10) all finalized normally, which requires stop-visibility to drop when generation ends; none stalled to the response timeout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Browser: scope the fallback stop-control selector to the composer so read-aloud, dictation, and voice controls cannot hold completed responses open until timeout. Thanks @StartupBros!
 - Browser: support ChatGPT GPT-5.6's unified Intelligence picker, where the menu wraps `composer-intelligence-picker-content` and the highest effort is labeled `Pro` instead of `Pro Extended`; recognize the current Chinese effort labels (`极速5.5`, `中`, `高`, and `极高`) without prefix collisions and verify switches against React-replaced composer pills. Fixes #303. Thanks @DragonFSKY!
 - Browser: keep hidden macOS Chrome windows rendered off-screen so trusted prompt submissions land without retaining drafts or leaking them into later runs. Fixes #298 and #312. Thanks @LeoLin990405!
 - Browser: require positive terminal evidence before finalizing ChatGPT responses so settled preambles and mid-stream text cannot be captured as the completed answer. Thanks @StartupBros!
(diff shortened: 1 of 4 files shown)


StartupBros:
Added the requested redacted real-behavior proof to the PR body: a live CDP capture during an active generation showing the real stop control (aria-label "Stop answering") inside the composer form, matched 1:1 by the scoped fallback, plus post-deploy production runs finalizing normally. @clawsweeper re-review

PR #318, opened 2026-07-12
fix(browser): report verified model labels in CLI output

### LINKED ISSUE #317: Browser: distinguish requested model keys from verified picker labels
## Summary

Browser-mode CLI output currently uses the requested/internal model key in places that read as though they describe the model active in ChatGPT. These values can differ because browser aliases are mapped to picker labels and because `current`/`ignore` strategies do not necessarily select the requested target.

GPT-5.6 makes the distinction especially visible:

- `gpt-5.6` currently targets the `GPT-5.6 Sol` picker entry.
- `gpt-5.5-pro` and older Pro aliases target ChatGPT's generic `Pro` entry.
- A verified bare `Pro` picker result proves the selected tier, but not a server-side GPT version, so it must not be expanded to `GPT-5.6 Pro`.

## Current behavior

For example:

```sh
oracle --engine browser --model gpt-5.6 --dry-run summary \
  -p "Reply exactly MODEL_DISPLAY_SMOKE_OK."
```

The launch summary identifies the run only as `gpt-5.6`. Completed-session summaries and status rows likewise use `metadata.model`, even when persisted model-selection evidence contains a different verified `resolvedLabel`.

This is also misleading with `--browser-model-strategy current`: the current picker label may differ from the requested key, while the compact completion summary still prints the requested key.

## Expected behavior

- Before browser selection, label the values as a target and requested key rather than presenting either as observed fact; for example, `target=GPT-5.6 Sol; requested=gpt-5.6`.
- Persist and print the internal key alongside the existing picker evidence.
- After selection, use the exact `resolvedLabel` in session/status/completion displays only when `verified=yes`.
- Keep the requested/internal key when evidence is absent or unverified.
- Never infer a concrete GPT version from the generic `Pro` picker label.

## Scope

This is a display/provenance issue, not a GPT-5.6 picker-selection failure. API-mode output should remain unchanged.

Observed on current `main` at `575e0881ca28beb76db6460a84a28405aaedb117` after #314.

DragonFSKY:
## Summary

- Distinguish browser picker targets from requested CLI model keys before selection.
- Use exact picker labels in CLI and TUI session displays only when persisted evidence is verified.
- Preserve the generic `Pro` picker label without inferring a server-side GPT version.
- Keep raw model keys for API output, model filtering, per-model log routing, and non-primary model runs.

Fixes #317.

## Display contract

Browser model identity has three separate sources:

- `requestedKey`: the Oracle model key, such as `gpt-5.5-pro`, `gpt-5.6`, or `gpt-5.6-sol`.
- `target`: the label Oracle asks the ChatGPT picker to select, such as `Pro` or `GPT-5.6 Sol`.
- `resolvedLabel`: an observed picker/model label after selection. It becomes the effective display label only when it is non-empty and `verified=true`.

Before selection, launch and dry-run output reports intent rather than observed state:

```text
Launching browser mode (target=GPT-5.6 Sol; requested=gpt-5.6) ...
Launching browser mode (picker=current; requested=gpt-5.5-pro) ...
```

After selection, compact status and completion output uses the exact verified label, otherwise it falls back to the requested key. Detailed CLI and TUI views retain both when they differ:

```text
Pro (requested gpt-5.5-pro)
```

The two Sol aliases retain distinct requested keys even though they currently share the `GPT-5.6 Sol` target. A bare `Pro` result remains `Pro`; no path expands it to `GPT-5.6 Pro`.

## Evidence hardening

- A picker target is never copied into `resolvedLabel`. If the UI exposes only effort text or no usable label, the stored value is `null`.
- `verified=yes` now means that a non-empty observed label was available after an explicit selection strategy; a successful selection without observable label evidence remains `verified=no`.
- The live browser finish line uses the same verified-label display helper as stored-session summaries, status tables, and TUI headers.
- The browser status still records `switched` or `already-selected` separately, so the lack of a displayable label does not overwrite selection state.

## Implementation notes

- Shared browser-layer formatters avoid a browser-to-CLI dependency.
- Live and persisted evidence use one stable field order: `requestedKey`, `target`, `resolvedLabel`, `status`, `strategy`, `verified`, `source`, and `capturedAt`.
- Session-level picker evidence applies only to the primary `metadata.model`; other model runs keep their raw keys.
- Legacy browser sessions using `options.mode` receive the same display behavior.
- TUI action values continue to use raw keys so per-model log lookup is unchanged.
- API-mode output and routing are unchanged.

## Validation

- `pnpm check`
- `pnpm build`
- Focused picker, runner, display, status-table, TUI, and integration tests: 181 passed
- `env -u ANTHROPIC_BASE_URL pnpm vitest run`: 1,570 passed, 44 skipped across 164 files

The focused coverage includes the two provenance regressions fixed here:

- a successful picker operation without an observable label cannot produce `resolvedLabel=Pro; verified=yes` from the requested target;
- a verified `Pro` label appears in the live browser finish line, while unverified evidence retains the requested key.

### Real persisted-browser evidence

A completed signed-in browser session stored `requestedKey=gpt-5.5-pro`, `target=Pro`, `resolvedLabel=Pro`, and `verified=yes`. Replaying that real metadata through the PR produced:

```text
Recent Sessions
Status    Model
completed Pro

Models:
- Pro (requested gpt-5.5-pro) — completed
Browser evidence:
- model requestedKey=gpt-5.5-pro; target=Pro; resolvedLabel=Pro; status=already-selected; strategy=select; verified=yes; source=chatgpt-model-picker
```

The interactive TUI for the same persisted session showed:

```text
Model: Pro
Models:
 - Pro (requested gpt-5.5-pro) — completed
```

The TUI log action still used `gpt-5.5-pro`, proving display formatting does not alter log routing. Local paths, conversation identifiers, and browser profile details are omitted.


DragonFSKY:
@clawsweeper re-review

Addressed the TUI consistency finding, added strategy-aware pre-selection provenance, prevented session evidence from leaking across model runs, unified live/persisted evidence formatting, and added redacted real persisted-browser replay output to the PR description.

DragonFSKY:
@clawsweeper re-review

Addressed the evidence-provenance and live finish-line findings in `72a17e83a88df138ed1f9dcbf5424d70320e2be7`:

- `resolvedLabel` is now observed-only; picker targets and effort-only labels cannot become verified display evidence.
- The live browser finish line now uses the verified display label and falls back to the requested key for unverified evidence.
- Added focused regressions for both boundaries.

Validation: `pnpm check`, `pnpm build`, focused 181 tests, and full suite 1,547 passed / 44 skipped. All GitHub CI and GitGuardian checks are green.

steipete:
Maintainer repair pushed in `f1d12438`.

The no-label success path now records GPT-5.6 Sol as explicitly unverified instead of running strict label validation against an empty string and throwing. Positive observed labels remain strictly validated.

Proof:

- 144 focused picker/display/session tests passed
- built-library invocation returned `status=already-selected`, `resolvedLabel=null`, and `verified=false` for `gpt-5.6-sol`

Recommendation: **LAND** once refreshed checks complete.

steipete:
Final reconciliation pushed in `ea47e7ce`: current `main` is merged, and the #317 release note now sits under `0.16.2 — Unreleased` with @DragonFSKY credit instead of the shipped 0.16.0 section. The refreshed branch passes 149 focused tests, check, build, and autoreview.

PR #319, opened 2026-07-12
Discover live model inventory + data-driven picker matching (robust to new models)

xuan-wei:
## Problem

Browser-mode model selection is hard-coded in three layers:

1. `cli/options.ts` `inferModelFromLabel` — text → known model id
2. `cli/browserConfig.ts` `BROWSER_MODEL_LABELS` — model id → UI label
3. `browser/actions/modelSelection.ts` `scoreOption` — a per-version ladder (`desiredVersion`, `versionFromLabel`, per-version `includes` guards, `data-testid` tokens)

Every new ChatGPT model needs edits in all three. GPT-5.6 "Sol" landed in `main` as ~10 hand-added `'5-6'` tokens. On published `0.15.2` (no 5.6 tokens), `-m "GPT-5.6 Sol"` silently collapses to `gpt-5.2` and fails:

```
[retry] Unable to find model option matching "GPT-5.2" in the model switcher.
Available: Instant5.5, Medium, High, Extra High, Pro, GPT-5.6 Sol.
```

Notably, a live capture shows the GPT-5.6-era picker **dropped every `data-testid`** — it's now a clean version × effort matrix keyed on `role` + text + `aria-checked`, which is exactly what makes runtime discovery reliable.

## Approach — Discover → Match → Apply → Verify

Read what the picker **actually offers**, then match against it, instead of precomputing a rigid target from a hard-coded ladder:

- **Discover** (`enumerateModelInventory`) — scrape the live menu (incl. the version submenu) into a structured `ModelInventory` (`versions`, `efforts`, current markers).
- **Match** (`matchModelToInventory`) — parse the request generically (any version number, effort synonyms) and resolve it against the live options.
- **Apply / Verify** — click the chosen option(s) by exact text; verify from the top-menu labels.

Because version parsing is generic, a brand-new model (`gpt-5.7`, `o4`, …) resolves the moment ChatGPT lists it — **no code change**. Unavailable requests fail loud with the real candidate list.

**Zero-regression:** the inventory path runs first for `strategy: "select"`; on any discovery/match/apply/verify miss it returns `null` and falls through to the untouched legacy `buildModelSelectionExpression`. Worst case is one extra discovery round-trip, then exactly today's behavior.

## Live verification (against a signed-in ChatGPT, GPT-5.6 "Sol", 2026-07)

| `-m` input | result |
|---|---|
| `"GPT-5.6 Sol"` | `Model picker (inventory): GPT-5.6 Sol` — the input that fails on 0.15.2 |
| `gpt-5.5-pro` | `Model picker (inventory): Pro GPT-5.5` — real version switch (5.6 Sol → 5.5) |
| `gpt-5.4` | `Model picker (inventory): GPT-5.4` — switch; "Leaving on July 23" note stripped |
| `o3` | `Model picker (inventory): o3` — a model the CLI can't name, selected from the live menu |
| `gpt-9.9` (invalid) | `Model inventory: "gpt-9.9" → not-found (available: GPT-5.6 Sol, GPT-5.5, GPT-5.4, GPT-5.3, o3); using legacy picker.` |

The raw `-m` value is threaded as `modelRequest` (not the CLI-inferred id), so `o3` / future models resolve from the live menu instead of being pre-collapsed to `gpt-5.2`.

Two DOM quirks found + handled during live testing (covered by tests):
- The composer pill merges version+effort for non-default versions (`5.5Pro`), so verification reads the top-menu **version-trigger label** + **checked effort radio**, not the pill.
- After a switch, a **spurious empty** `aria-haspopup="menu"` item can precede the real version trigger — trigger detection now requires version-like text.

## Tests

- 26 new unit tests, incl. the 0.15.2 failure cases, a **future-model** case, and a `new Function()` parse-check of every generated browser expression.
- Pure classifier/matcher extracted to `src/oracle/` (unit-tested); browser DOM ops in `src/browser/actions/modelInventoryScrape.ts`.
- `tsc --noEmit` clean; existing `browserConfig` / `browserDefaults` tests still pass.

## Scope / follow-ups (separate PRs)

- Optional **LLM fallback** for `ambiguous` / `not-found`, with the enumerated options as an **enum constraint** (can't hallucinate), gated behind API-key availability / opt-in, cached, off by default.
- Once discovery is proven in the field, retire the hard-coded version ladder and keep it only as a deep fallback.

Happy to adjust the direction — this is intentionally additive and behind a fallback so it can land incrementally. See `docs/model-discovery-proposal.md` for the full design.

### DIFF
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -2113,6 +2113,7 @@ async function runRootCommand(options: CliOptions): Promise<void> {
       ...options,
       remoteHost: remoteHost ?? undefined,
       model: activeModel,
+      modelRequest: cliModelArg,
       browserModelLabel: resolveBrowserModelLabel(cliModelArg, activeModel),
     });
     return resolvedOptions.browserResumeConversationUrl
--- a/docs/model-discovery-proposal.md
+++ b/docs/model-discovery-proposal.md
@@ -69,10 +69,16 @@ miss, the existing `buildModelSelectionExpression` loop.
 
 Driven against a signed-in ChatGPT via the CLI running from source:
 
-- `-m "GPT-5.6 Sol"` → `Model picker (inventory): GPT-5.6 Sol` — the exact input that collapses to
-  `gpt-5.2` and fails on published 0.15.2 now resolves correctly (already-selected short-circuit).
-- `-m gpt-5.5-pro` → `Model picker (inventory): Pro GPT-5.5` — a real version switch (5.6 Sol → 5.5),
-  applied via the version submenu and verified.
+| `-m` input | result |
+|------------|--------|
+| `"GPT-5.6 Sol"` | `Model picker (inventory): GPT-5.6 Sol` — the input that collapses to `gpt-5.2` and fails on 0.15.2 |
+| `gpt-5.5-pro` | `Model picker (inventory): Pro GPT-5.5` — real version switch (5.6 Sol → 5.5) |
+| `gpt-5.4` | `Model picker (inventory): GPT-5.4` — switch; the "Leaving on July 23" note is stripped |
+| `o3` | `Model picker (inventory): o3` — a model the CLI can't name, selected purely from the live menu |
+| `gpt-9.9` (invalid) | `Model inventory: "gpt-9.9" → not-found (available: GPT-5.6 Sol, GPT-5.5, GPT-5.4, GPT-5.3, o3); using legacy picker.` |
+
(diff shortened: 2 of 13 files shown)


steipete:
Closing per the maintainer review and the detailed analysis already posted above. The inventory keys models primarily by numeric version, which loses variant identity when multiple targets share a version and can verify the wrong same-version picker option. The branch also overlaps newer, evidence-backed model-selection work now on main. The core abstraction is therefore unsafe for the intended automatic matching behavior.

Thank you for the investigation and contribution.

PR #321, opened 2026-07-12
feat(codex): add security findings workflow

umgbhalla:
## Motivation

Oracle previously had no Codex Cloud security-findings workflow. This PR adds the complete path from discovery to action: it can enumerate findings, open a selected finding, extract the report and evidence needed to understand it, and then invoke the finding’s supported detail-panel actions from the CLI.

The first part is deliberately read-only. It turns the authenticated findings surface into structured CLI output instead of forcing an operator to navigate the web UI, while preserving the report sections and evidence links that explain why a finding exists. The second part adds the operational step that was missing before: an explicit finding action command that can create a PR, start a finding chat, close or adjust a finding, and copy the available patch or report material.

### Discovery and report extraction

- `oracle codex findings` enumerates the paginated findings surface, reports severity and repository metadata, supports severity/limit filtering, and can emit JSON.
- `oracle codex findings --finding <selection-id>` opens one finding and extracts its title, repository/commit reference, Summary, Validation, Evidence, and Attack-path analysis sections.
- Detail extraction also returns deduplicated GitHub evidence links and the signed validation-artifact URL without fetching or modifying the artifact.
- Finding IDs are validated before they are used to build detail URLs, and the browser runner uses the authenticated browser surface rather than an unauthenticated API shortcut.

### Explicit finding actions

- `oracle codex finding --finding <selection-id> --action <action>` provides the action layer for `create-pr`, `chat`, `close`, `adjust`, `copy-content`, `copy-link`, `copy-patch`, and `copy-git-apply`.
- The action runner targets the accessible detail-panel controls with trusted CDP mouse events. This is required because the Radix pointer-event controls can ignore synthetic `element.click()` calls.
- Chat requires `--text`, while mutating actions require `--confirm`; read-only copy actions do not require confirmation.
- Create PR opens the git action menu and checks for `View PR` before clicking Create PR, so rerunning the command returns an existing PR instead of creating a duplicate.
- Actions can be constrained to an expected GitHub repository with `--repo`, and the finding detail is re-read immediately before the action is dispatched.

## Compatibility and safety

The original `oracle codex findings` command remains read-only, and adding the action command does not make list or detail inspection mutating. The runner refuses mutating actions without explicit confirmation and refuses to proceed when trusted browser input is unavailable, so a failed automation primitive cannot be reported as a successful action.

## Validation

- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run tests/codex/findings.test.ts` — 16 passed
- `pnpm run format:check`
- `pnpm run lint`
- `pnpm run build`
- `pnpm test` — 1,572 passed, 44 skipped

## Exact-head live proof

Current head: `1b7d6c7cfe42dc2668ccfde8c7f6c57b51b1cbc7`

- Authenticated findings launch returned 33 findings.
- A selected finding’s report was read successfully, including its evidence links and validation artifact reference.
- A trusted copy action completed successfully.
- Repeating Create PR returned that existing PR instead of creating a duplicate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Codex Cloud: add explicit `oracle codex finding --action ...` commands for trusted Create PR, Chat, Close, Adjust, and copy actions while keeping `oracle codex findings` read-only.
 
+### Fixed
+
+- Codex Cloud: scope finding actions to the detail panel, verify action-specific UI outcomes, enforce `--repo` consistently, and replace repository-specific evidence filtering with configurable include/exclude path policies.
+
 ## 0.16.0 — 2026-07-12
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.16.1 — Unreleased
 
+### Added
+
+- Codex Cloud: add explicit `oracle codex finding --action ...` commands for trusted Create PR, Chat, Close, Adjust, and copy actions while keeping `oracle codex findings` read-only.
+
 ## 0.16.0 — 2026-07-12
 
 ### Added
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -1024,7 +1024,7 @@ addProjectSourcesCommonOptions(
 
 const codexCommand = program
   .command("codex")
-  .description("Read-only access to ChatGPT Codex Cloud (security findings).");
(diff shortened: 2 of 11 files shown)


steipete:
Closing per the maintainer review and the detailed analysis already posted above. This is a broad Codex findings workflow with page-global PR/copy interactions and mutation of settings, but the proof does not establish that those operations remain scoped to the intended finding and repository. The resulting safety and verification surface is too large for the behavior delivered, so this is not suitable to land.

Thank you for the contribution.

PR #323, opened 2026-07-12
fix(cli): preserve current browser effort

jung0han:
## Summary

- do not inherit `browser.thinkingTime` from config when the CLI explicitly requests `--browser-model-strategy current`
- preserve an explicit `--browser-thinking-time` and config-defined `modelStrategy: current` plus `thinkingTime`
- document the precedence rule and cover it with direct and real Commander source tests

## Reproduction

With a global browser default such as:

```json
{
  "browser": {
    "thinkingTime": "extended"
  }
}
```

this command intends to keep the current browser state:

```bash
oracle --engine browser --browser-model-strategy current -p "..." --file "..."
```

Before this change, the config default was still injected, so a currently selected Pro model could fail before submission while Oracle tried to confirm a separate Extended effort chip.

The new precedence rule suppresses only that inherited value. Users can still change effort explicitly with `--browser-thinking-time`, and config-defined strategy/effort pairs remain unchanged.

## Verification

- `pnpm vitest run tests/cli/browserDefaults.test.ts` — 18 passed
- `pnpm vitest run` — 144 files passed, 18 skipped; 1,563 tests passed, 43 skipped
- `pnpm check`
- `pnpm build`
- signed-in Chrome smoke with config `thinkingTime: extended`, explicit `--browser-model-strategy current`, and no thinking-time flag
  - no thinking-time selection step ran
  - request completed with exact marker `CURRENT MODEL NO INHERITED EFFORT OK`
- Oracle external review after requested revisions
  - `BLOCKERS: None`
  - `TEST_GAPS: None material`
  - `VERDICT: APPROVE`

This is intentionally separate from #320: it changes CLI config precedence for the `current` strategy rather than GPT-5.6 Sol/Pro picker compatibility.


jung0han:
Oracle independent review completed against head `e827942f`.

- **BLOCKERS:** None
- **TEST_GAPS:** No material test gaps
- **VERDICT:** **APPROVE**

Oracle confirmed the source-sensitive guard preserves the intended matrix: explicit CLI `current` suppresses only inherited config thinking time; explicit CLI thinking time is retained; config-defined `current + thinkingTime` remains intact; and explicit `select`/`ignore` still inherit the config value.

Non-blocking suggestions only: optionally spell out `select`/`ignore` inheritance in docs and add a future root-command config integration test. The review considered the current focused/full tests, cross-platform CI, build/check, and signed-in browser smoke sufficient for merge readiness.

PR #324, opened 2026-07-13
fix(browser): ignore loopback WSL resolver hosts for launched Chrome

Rokurolize:
## Problem

On WSL, launcher mode uses the first `/etc/resolv.conf` nameserver as the Chrome DevTools connection host. With a distro-managed systemd-resolved stub, that value is a guest-local DNS endpoint such as `127.0.0.53`, not a Windows host address. Oracle therefore launches Chrome but checks the dynamic DevTools port through the DNS stub and fails before browser automation begins.

The same resolver heuristic was duplicated in the browser connectivity smoke script.

## Why `127.0.0.53` can appear on WSL

This address is not assigned by WSL as the Windows host. It appears when Linux DNS management is layered on top of WSL networking:

1. [Microsoft documents that WSL supports systemd](https://learn.microsoft.com/en-us/windows/wsl/systemd), and current Ubuntu installations made through `wsl --install` use systemd by default. A WSL distribution can therefore run `systemd-resolved` just like a regular Linux installation.
2. [Microsoft documents that `generateResolvConf` controls whether WSL generates `/etc/resolv.conf`](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#network-settings). The reproduced distribution explicitly has `systemd=true` and `generateResolvConf=false`, so `/etc/resolv.conf` is managed inside the distribution and points to systemd-resolved's stub file.
3. [systemd's official `resolved.conf` manual](https://www.freedesktop.org/software/systemd/man/latest/resolved.conf.html#DNSStubListener=) defines `127.0.0.53` and `127.0.0.54`, port 53, as local DNS stub listener addresses. The full resolver is exposed on `127.0.0.53`; it forwards queries to the currently selected upstream DNS server.
4. [RFC 1122 section 3.2.1.3](https://www.rfc-editor.org/rfc/rfc1122#section-3.2.1.3) defines every address of the form `127.<any>` as internal host loopback that must not appear outside a host. A resolver-derived `127.0.0.53` therefore cannot identify a separate Windows host across the WSL network boundary.
5. [Microsoft's WSL networking guidance](https://learn.microsoft.com/en-us/windows/wsl/networking#identify-ip-address) identifies the Windows host in NAT mode through the default route and notes that mirrored mode uses `localhost`. It does not define a `127/8` address other than local loopback as a remote Windows-host endpoint.
6. [Microsoft's WSL configuration reference](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#experimental-settings) separately documents `10.255.255.254` as the default `dnsTunnelingIpAddress`. On the reproduced host, systemd-resolved uses that address as its upstream DNS server while exposing `127.0.0.53` to local Linux clients.

The observed resolver chain was:

```text
/etc/resolv.conf -> /run/systemd/resolve/stub-resolv.conf
nameserver 127.0.0.53
systemd-resolved upstream DNS server: 10.255.255.254
```

The old heuristic only works when the first nameserver is a reachable non-loopback host-side address. It becomes invalid when `/etc/resolv.conf` names a local resolver frontend. Oracle does not need to infer the complete WSL topology for this fix: rejecting resolver-derived `127/8` values is sufficient because those values are unambiguously guest loopback.

## Change

This centralizes WSL host resolution, maps only resolver-derived `127/8` values to `127.0.0.1`, and uses the standard local launcher for that route. Explicit `ORACLE_BROWSER_REMOTE_DEBUG_HOST` and `WSL_HOST_IP` values remain authoritative. Resolver-derived non-loopback values retain the patched remote-host launcher and `0.0.0.0` bind behavior for Windows Chrome compatibility.

The runtime and `scripts/test-browser.ts` now use the same resolver parser. The WSL behavior is recorded in `docs/windows-work.md` and the unreleased changelog.

## Reproduction and live proof

The original comparison used upstream `8e3456e7` and submitted head `527857f1` on the same WSL2 host. Both host overrides were removed with `env -u ORACLE_BROWSER_REMOTE_DEBUG_HOST -u WSL_HOST_IP`.

Upstream, using a fixed diagnostic port:

```text
EXPECTED_FAILURE connect ECONNREFUSED 127.0.0.53:45903
```

After rebasing onto `19a0b79b`, I repeated the production `launchChrome` path on rebased head `1cd2abb8` with the same overrides unset. The live resolver was the systemd-resolved stub:

```text
proof_head=1cd2abb8ef8e5aedb0bd21f1a09cda98aa0a9a63
resolver_host=127.0.0.53
route={"connectHost":"127.0.0.1","debugBindAddress":"127.0.0.1","usePatchedLauncher":false}
Launched Chrome (pid redacted) on port 45904 on 127.0.0.1
PATCHED_SUCCESS host=127.0.0.1 port=45904 status=200
```

The success status is from Chrome's live `/json/version` endpoint. The failed upstream launch was terminated after capture.

## Validation

On rebased head `1cd2abb8`, `pnpm vitest run tests/browser/wslHost.test.ts tests/browser/chromeLifecycle.test.ts` passes 35 tests. Coverage includes `127.0.0.53`, `127.0.0.54`, another `127/8` value, non-loopback preservation, explicit override precedence, explicit loopback override preservation, legacy override preservation, and local versus patched launcher routing.

`pnpm check`, `pnpm build`, and `git diff --check` pass.

`pnpm test` passes 1,601 tests in 147 files, with the existing 43 tests and 18 files skipped.

## Non-goal

This does not redesign the broader WSL topology heuristic or change attach-running and explicit remote-Chrome behavior. Distinguishing native Linux Chrome, Windows Chrome, and remote Chrome using positive topology evidence remains separate work.


Rokurolize:
@clawsweeper re-review

steipete:
Final reconciliation pushed in `7e72ffcc`: current `main` is merged, and the WSL fix release note now sits under `0.16.2 — Unreleased` with @Rokurolize credit instead of the shipped 0.16.1 section. The refreshed branch passes 42 focused browser/WSL-adjacent tests, check, build, and autoreview.

PR #332, opened 2026-07-21
fix(gemini): type mp4/mov/webm uploads so Gemini actually receives them

mkubenka:
## The bug

`GEMINI_UPLOAD_MIME_TYPES` covers only images and PDF. Everything else falls through to `application/octet-stream`, and **Gemini silently discards uploads it cannot type**.

The failure is invisible from the outside, which is what makes it expensive:

```
$ oracle --engine browser -m gemini-3.1-pro --browser-attachments always \
    --file clip.mp4 -p "Describe this video. If you did not receive a video, say NO VIDEO RECEIVED."

Answer:
NO VIDEO RECEIVED.

7.4s · gemini-3.1-pro[browser] · ↑51 ↓5
files=1        <-- reports the file as sent
```

The upload succeeds, the payload is built, `files=1` is printed — and the model gets nothing. Reading the source reinforces the wrong conclusion, because the upload path genuinely is complete and correct. Only the MIME label is wrong.

Declaring `video/mp4` makes the same file work on the first attempt:

```
Answer:
The video features an interview setup in a studio environment.
* **Visuals:** A man wearing glasses and a black shirt sits in front of a microphone...
  Chinese and English subtitles appear at the bottom of the screen...
### Verbatim English Subtitles (First 15 Seconds):
* "so it snows."
```

## The change

Adds exactly three entries — `.mp4`, `.mov`, `.webm` — and extracts the lookup into an exported `resolveGeminiUploadMimeType()` so it is unit-testable. No behaviour change for currently-mapped types; the `application/octet-stream` fallback is preserved.

## Why only three

I first wrote a much larger table from the Gemini API's documented formats, then probed each candidate against the live endpoint with a sentinel prompt separating "received" from "actually viewable". Most of it was wrong:

| Format | Result |
|---|---|
| `.mp4` `.mov` `.webm` | **viewable** — describes people, clothing, on-screen text |
| `.avi` | received but **not viewable**; only echoed container metadata (`Lavf62.3.100`) |
| `.m4v` | **dropped**, despite byte-identical content to a working `.mp4` and the same declared `video/mp4` |
| `.mkv` `.3gp` `.heic` | Gemini returns an error |
| `.wav` `.mp3` `.m4a` | never delivered — tried `audio/mpeg`, `audio/mp3`, `audio/wav`, `audio/x-wav` |
| `.txt` | never delivered — tried `text/plain` and `text/md` |

Two conclusions worth recording:

1. **Audio and plain text are not supported by this endpoint at all**, whatever type is declared. Mapping them would advertise support that does not exist.
2. **The endpoint gates on the file extension, not just the declared MIME type.** The `.m4v` case is the proof: same bytes, same `video/mp4`, still dropped. So this table cannot be extrapolated from the Gemini API docs — each entry needs probing.

I have kept the entries to what I verified. If you have a broader whitelist from the web app, more can be added safely.

## Verification

Against `gemini-3.1-pro`:

- 1.7 MB / 60 s MP4 — before: `NO VIDEO RECEIVED`; after: accurate scene description plus verbatim on-screen subtitles, cross-checked against frames read locally.
- A full 99-minute video in ten ~8.6 MB chunks (`scale=640:-2,fps=8`, `crf 32`). Answers to timestamp-specific questions matched an independently produced OCR transcript, including subtitles at 5:00 and 5:48 and both speakers' names.

Checks: `vitest` (5 new tests plus existing `tests/gemini.test.ts` — 15 passing), `tsc --noEmit`, `oxfmt --check`, `oxlint`.

## Notes for the maintainer

- **The 1 MB `DEFAULT_MAX_FILE_SIZE_BYTES` still applies**, so video realistically needs `--max-file-size-bytes` / `ORACLE_MAX_FILE_SIZE_BYTES`. Left alone as a separate policy decision.
- `uploadGeminiFile()` reads the whole file into a Buffer→Blob, so large media will exhaust the heap. Chunking is the practical answer; out of scope here.
- **The deeper issue is the silent failure**, not the missing entries. An extension falling back to `application/octet-stream` arguably deserves a warning, since the current behaviour reports success while the model receives nothing — that is what made video look impossible rather than merely broken. There is no logger plumbed into `client.ts`'s upload path, so I kept this PR to the data fix. Happy to follow up if you want the warning.


mkubenka:
@clawsweeper re-review

steipete:
Maintainer repair pushed through `c94c844e`.

I reconciled the release note to the current `0.16.2 — Unreleased` section, kept it to house style with contributor credit, and merged current `main` without rewriting Michal's original commit.

Proof on the resulting branch:

- 5 focused MIME resolver tests passed
- built-library smoke returned `video/mp4`, `video/quicktime`, and `video/webm`, while `.m4v` remained `application/octet-stream`

Recommendation: **LAND** once the refreshed checks complete.

PR #334, opened 2026-07-22
fix(browser): wait for durable ChatGPT conversation URLs

### LINKED ISSUE #333: Browser: completed ChatGPT response is never captured when thinking indicator is absent
## Summary

On current ChatGPT browser UI, Oracle can successfully submit a small text-only prompt and ChatGPT can complete the answer, while Oracle continues reporting that no thinking status was detected and never captures the completed assistant turn.

This appears to be a regression or remaining gap after #284 / #293. It is distinct from #327: no DevTools disconnect occurred in this reproduction.

## Reproduction

1. Run Oracle in browser mode through `oracle serve` against an already signed-in, persistent manual-login Chrome profile.
2. Use the current active model (`gpt-5.5-pro` here) and submit a short synthetic exact-token prompt; no attachments are needed.
3. Wait for the browser response timeout.

## Actual behavior

Oracle repeatedly logged:

```text
Waiting for ChatGPT response - <elapsed>; no thinking status detected yet.
```

The bounded foreground client timed out without a captured answer. It did not report a CDP disconnect.

Direct CDP inspection of the still-live ChatGPT conversation showed:

- a `/c/<id>` conversation URL;
- the synthetic exact token appeared twice in `document.body.innerText` (the submitted user turn and the completed assistant reply); and
- no visible `data-testid="stop-button"` control.

So ChatGPT had completed the response, but Oracle did not recognize or harvest it.

## Expected behavior

Once Oracle has submission evidence and a conversation URL, lack of a recognized thinking indicator should not by itself prevent response capture. Oracle should fall back to polling the current conversation for a newly added assistant turn and require robust terminal evidence before completing it.

The existing safeguards from #293 still matter: a fallback should use a post-submit turn baseline, associate the assistant turn with the current conversation, and require stable content plus no active streaming control before accepting a result.

## Environment

- Oracle published `0.16.0`, then current `main` at `f13347973b` (the unreleased 0.16.1 changelog changes were installed for confirmation)
- Linux x86_64, Google Chrome 144.0.7559.109
- Browser engine, remote `oracle serve`, persistent manual-login profile
- Current-model strategy; active model reported as `gpt-5.5-pro`

## Related

- #284 identified missing thinking-state detection and stale-tab recovery; it was closed by later fixes, but this short current-UI reproduction still misses a completed assistant turn.
- #293 added stricter completion proof, which a fallback should preserve rather than bypass.
- #327 is installed in the reproduction and is not exercised here because Chrome and the target stayed connected.

I can test a focused patch against the same Linux remote-service setup.

dbachko:
## Summary

- ignore ChatGPT's transient `/c/WEB:<request-id>` route until the durable conversation URL appears
- centralize stable conversation URL extraction across response capture, live-tab harvesting, and reattachment
- cover the transient-to-durable navigation sequence with regression tests

## Root cause

ChatGPT can briefly navigate to `/c/WEB:<request-id>` after submission before replacing it with `/c/<conversation-uuid>`. Oracle's loose conversation URL matcher accepted the transient route and persisted `WEB` as the expected conversation ID. The completed assistant turn belonged to the durable UUID, so response capture rejected it as out of scope and continued polling until timeout even though the answer was visible and terminal.

The shared extractor now requires the conversation path segment to end before `/`, `?`, or `#`, which rejects the colon-delimited transient route. Monitoring continues until the durable URL is available, and reattach/live-tab helpers use the same rule.

This runtime fix is separate from #318, which corrects display and provenance for the bare `Pro` picker label.

## Validation

- `pnpm check`
- focused browser suite: 303 passed, 1 skipped
- full suite: 1,591 passed, 43 skipped
- live GPT-5.6 Sol browser run: explicitly selected and verified, completed in 13.5 seconds
- live Pro browser run: explicitly selected and verified, completed in 16.4 seconds
- live Pro Deep Research run: completed in 2 minutes 47 seconds and reattached with validated report/transcript artifacts
- pre-fix stuck GPT-5.6 Sol session: reattached and marked completed with the patched URL scope

No live test clicked ChatGPT's **Answer now** control.

Fixes #333.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Browser: ignore transient `/c/WEB:<request-id>` routes until ChatGPT exposes the durable conversation URL, preventing completed GPT-5.6 and Pro answers from hanging until timeout under a mismatched response scope. Fixes #333. Thanks @kesslerio!
+- Browser: ignore transient `/c/WEB:<request-id>` routes until ChatGPT exposes the durable conversation URL, preventing completed GPT-5.6 and Pro answers from hanging until timeout under a mismatched response scope. Fixes #333. Thanks @dbachko and @kesslerio!
 - Browser: recover completed answers after a recoverable DevTools disconnect by confirming target liveness and attempting bounded reattachment, while preserving fail-closed handling for unavailable targets. Fixes #326. Thanks @piyushbag!
 - CLI: avoid inheriting `browser.thinkingTime` from config when `--browser-model-strategy current` is explicit, while preserving an explicit `--browser-thinking-time` override. Thanks @jung0han!
 - Browser/Serve: keep the authenticated manual-login Chrome process alive while closing each successfully captured service-owned run tab, preventing renderer and memory accumulation across repeated remote consultations without changing explicit `--browser-keep-browser`, attached-tab, or incomplete-run recovery behavior. Thanks @rtl-ai!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Browser: ignore transient `/c/WEB:<request-id>` routes until ChatGPT exposes the durable conversation URL, preventing completed GPT-5.6 and Pro answers from hanging until timeout under a mismatched response scope. Fixes #333. Thanks @kesslerio!
 - Browser: recover completed answers after a recoverable DevTools disconnect by confirming target liveness and attempting bounded reattachment, while preserving fail-closed handling for unavailable targets. Fixes #326. Thanks @piyushbag!
 - CLI: avoid inheriting `browser.thinkingTime` from config when `--browser-model-strategy current` is explicit, while preserving an explicit `--browser-thinking-time` override. Thanks @jung0han!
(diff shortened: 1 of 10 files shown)


kesslerio:
Validated this against my real Linux manual-login remote-browser route, rather than direct CDP.

Before this patch, released Oracle 0.16.0 could submit a one-line prompt and save the completed ChatGPT answer in the local transcript, but then kept polling until it marked the session zombie/error.

With commit 2157ab73eb5445ebcfbb7ebe52fbf988352334d6 on the same persistent signed-in profile and the same oracle serve -> remote-client route:
- a server-route smoke completed and captured the expected answer in 28.0s;
- after pinning that build as my temporary local hotfix, the normal remote-client smoke completed and captured the expected answer in 15.3s.

Local validation:
- pnpm vitest run tests/browser/conversationUrlMonitor.test.ts tests/browser/reattach.test.ts — 21 passed
- pnpm run build — passed

This fixes the transient-conversation-URL capture failure I was seeing. I will keep the local hotfix until a release contains this commit.

kesslerio (commented):
P2 — `src/browser/reattachability.ts:24` remains outside this PR's shared stable-URL contract.

The broad `/(?:^|\/)c\/[^/]+/` check still accepts transient `/c/WEB:<request-id>` routes as recoverable. As a result, `hasRecoverableChatGptConversation()` can still approve stale metadata for automatic reopening, even though the new parser rejects that route in response capture and other reattachment helpers.

Please route this gate through `isStableConversationUrl()` / `extractStableConversationIdFromUrl()` and add a regression test proving a `WEB:` URL is not recoverable without a durable conversation id.

PR #340, opened 2026-07-26
fix(browser): recognize Sol Pro effort control

Rokurolize:
## Summary

ChatGPT's current GPT-5.6 Sol composer exposes `Pro` as a separate reasoning-effort choice inside its combined intelligence menu. This change lets Oracle recognize and select that control when `--browser-thinking-time heavy` requests the highest available effort.

## Problem

Oracle's browser engine runs JavaScript in the signed-in ChatGPT page to inspect the composer and choose a reasoning effort. Its GPT-5.6 path recognized the earlier `Extra high` label and a dedicated Pro-effort trigger, but the current UI can instead present a generic `Pro` item alongside a separate `GPT-5.6 Sol` model item in one menu.

In that layout, Oracle could report `option-not-found` even when Pro was already selected, or fail to select Pro when another effort was active. This is separate from model selection: open PR #320 teaches Oracle to recognize the `GPT-5.6 Sol` model item when the same menu also contains a Pro pill, while this PR handles the independent effort item after the model has been chosen.

## Implementation

- For GPT-5.6 targets at `heavy` effort only, treat a standalone `Pro` item in the intelligence-effort menu as the target effort.
- Accept the current composer pill as already selected when its normalized label contains the standalone `Pro` token.
- Exclude labels containing `GPT` or a `5.6` model version when searching for the effort item, so the code does not mistake a model label for an effort.
- Preserve the existing behavior for other models and effort levels.

The test fixture models the current combined menu with separate model and effort entries. It covers both an already-selected Pro pill and switching an unchecked Pro radio item.

## Live browser proof

I ran a signed-in ChatGPT browser smoke at commit `4086ae1298f0353ebc8649ce8656dcd1233095ce` with the current UI:

```text
node dist/bin/oracle-cli.js --engine browser --browser-manual-login --model gpt-5.6-sol --browser-model-strategy select --browser-thinking-time heavy --browser-archive never --browser-attachments never --wait --slug current-ui-sol-effort-smoke --prompt "Current ChatGPT UI model and effort compatibility smoke after the Sol/Pro-pill fixes. Return exactly ORACLE_SOL_HEAVY_OK and nothing else." --file src/cli/detach.ts
```

Relevant output:

```text
[browser] Thinking time: Pro (already selected)
[browser] Model selection evidence: requested=GPT-5.6 Sol; resolved=GPT-5.6 Sol; status=already-selected; strategy=select; verified=yes.
Answer:
ORACLE_SOL_HEAVY_OK
```

The run completed in 18.3 seconds. No `Answer now` control was clicked.

## Validation

- `pnpm check`
- `pnpm vitest run tests/browser/thinkingTime.test.ts`
- `pnpm run build`
- `pnpm run test:packed-cli`

## Compatibility and security

This change does not add a network endpoint, alter authentication, copy browser credentials, or change model selection. It only broadens recognition within the existing ChatGPT effort menu for the GPT-5.6 `heavy` mapping.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Browser: run long local Pro consultations in a detached worker while the CLI remains attached to its session log, so unexpected foreground termination cannot stop answer capture; Ctrl-C still cancels the worker. Thanks @Rokurolize!
+- Browser: recognize ChatGPT's separate Pro effort control as the selected maximum effort for GPT-5.6 Sol when `--browser-thinking-time heavy` is requested.
 
 ## 0.16.1 — 2026-07-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixed
 
 - Browser: run long local Pro consultations in a detached worker while the CLI remains attached to its session log, so unexpected foreground termination cannot stop answer capture; Ctrl-C still cancels the worker. Thanks @Rokurolize!
-- Browser: recognize ChatGPT's separate Pro effort control as the selected maximum effort for GPT-5.6 Sol when `--browser-thinking-time heavy` is requested.
+- Browser: recognize ChatGPT's separate Pro effort control as the selected maximum effort for GPT-5.6 Sol when `--browser-thinking-time heavy` is requested. Thanks @Rokurolize!
 
 ## 0.16.1 — 2026-07-23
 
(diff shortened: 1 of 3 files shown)


steipete:
Maintainer repair `a21f29c0` reorders GPT-5.6 Sol heavy-effort matching so the standalone Pro control is checked before generic Extra High tiers. The submitted fixture remains, and a second regression now covers a combined Extra High + Pro + GPT-5.6 Sol menu.

Proof: focused thinking-time tests 27/27, `pnpm check`, clean build, full suite 1,605 passed / 43 skipped, and branch autoreview clean. Independent live browser proof remains unavailable on this Mac because it has no ChatGPT cookies; the contributor's signed-in live proof remains documented in the PR body.

PR #335, opened 2026-07-23
feat(api): add GPT-5.6 Pro reasoning controls

enki:
## Summary

- add `--reasoning-mode standard|pro` for GPT-5.6 Responses API runs
- add `--reasoning-effort none|low|medium|high|xhigh|max` for the complete GPT-5.6 effort set
- send Pro + max as `reasoning: { mode: "pro", effort: "max" }` without inventing a model slug
- persist both controls across detached sessions and restarts, with Pro timeout/background defaults
- reject unsupported models, unsupported mode routes, browser usage, and fake Pro model slugs
- document the API/browser distinction and cover request, CLI, persistence, configuration, and live-smoke paths

## Why

GPT-5.6 Pro is a Responses API reasoning mode, not a separate model slug. Oracle previously exposed a fixed reasoning effort but had no CLI path for Pro mode or GPT-5.6's new `none` and `max` effort levels. Inputs such as `gpt-5.6-sol-pro` could also fall through model normalization instead of failing with actionable guidance.

Mode and effort are independent: `standard` and `pro` are the complete GPT-5.6 mode enum, while `none`, `low`, `medium`, `high`, `xhigh`, and `max` are the complete GPT-5.6 effort enum. The [OpenAI reasoning-mode guide](https://developers.openai.com/api/docs/guides/reasoning#reasoning-mode) documents this request shape.

## Validation

After rebasing onto `main` at `68b8c51b`:

- full suite: 1,625 passed, 43 opt-in tests skipped
- focused reasoning, detachment, engine, timeout, background, and model-resolution tests: 166 passed
- formatting, lint, and typecheck passed
- live smoke now reserves 25,000 output/reasoning tokens so Pro + max can complete instead of being constrained by a 64-token budget
- dedicated real OpenAI GPT-5.6 Sol Pro + max background smoke passed
- CI passed on Ubuntu, macOS, Windows, CDP disconnect proof, and GitGuardian

### Live API proof

Validated against the real OpenAI Responses API on rebased head `b291cdf4`:

```text
ORACLE_LIVE_TEST=1 pnpm vitest run tests/live/openai-live.test.ts \
  -t "gpt-5.6-sol Pro reasoning mode background flow eventually completes"

Test Files  1 passed (1)
Tests       1 passed | 4 skipped (5)
Duration    7.08s
```

The synthetic smoke completed through Oracle's background create/retrieve path with `model: "gpt-5.6-sol"` and `reasoning: { mode: "pro", effort: "max" }`. No API key or private runtime data is included.

PR #342, opened 2026-07-27
fix(browser): support GPT-5.6 Sol Pro as a selectable target

Learn2Solve:
## Problem

Requesting Sol's Pro effort is currently impossible. `gpt-5.6-sol-pro` is rejected by the variant parser (`Unknown GPT-5.6 browser variant "sol pro". Use gpt-5.6 or gpt-5.6-sol.`), and even when the account is *already* on Sol Pro, selection fails:

```
Unable to find model option matching "GPT-5.6 Sol" in the model switcher.
Available: Instant5.5, Medium, High, Extra High, Pro, GPT-5.6 Sol.
```

Note "GPT-5.6 Sol" is listed as available yet reported as not found.

## Why

Dumped the live picker DOM (ChatGPT, 2026-07-27):

| text | role | attrs |
|---|---|---|
| `Instant5.5` / `Medium` / `High` / `Extra High` / `Pro` | `menuitemradio` | `Pro` has `aria-checked=true` |
| `GPT-5.6 Sol` | `menuitem` | `aria-haspopup="menu"` |

So **Pro is an effort radio layered on the active model** - the same shape `gpt-5.5-pro` already maps to (`"Pro"`) - while `GPT-5.6 Sol` is a *submenu opener*, not a selectable model row.

With the account on Sol Pro the composer pill reads `Pro`, which trips:

- `canTrustSelectedOption`: `if (desiredVersion === '5-6') return !hasProComposerPill();`
- three sibling guards keyed on `desiredModelVariant === 'sol' && hasProComposerPill()`

Those are right for plain Sol and exactly backwards once Pro is what was asked for.

This is adjacent to #340 (which taught `thinkingTime: heavy` to recognise the Pro effort control) but distinct: that fix assumes Sol can already be selected, and the selection step itself is what fails here.

## Change

- add `gpt-5.6-sol-pro` to `ModelName`, mapping to browser label `"Pro"` and surviving `normalizeChatGptModelForBrowser`
- accept `sol pro` / `sol-pro` / `pro` in the GPT-5.6 label parser - Sol is the only 5.6 model with a Pro effort, so bare `gpt-5.6-pro` is unambiguous
- gate the four Pro-pill vetoes on `!wantsPro`; require the pill instead of forbidding it when Sol Pro was requested
- API has no Pro model id (`/v1/models/gpt-5.6-pro` and `/v1/models/gpt-5.6-sol-pro` both 404 `does not exist`, verified 2026-07-27), so API runs resolve to `gpt-5.6-sol` at its highest exposed effort

## Verification

Live ChatGPT Pro account, before vs after:

```
before: Unable to find model option matching "GPT-5.6 Sol" in the model switcher.
after:  Model selection evidence: requested=Pro; resolved=Pro;
        strategy=select; verified=yes
        Answer: GPT-5.6 Pro，高推理档位。
```

`verified=yes` is the point - previously the only way to reach Pro was `--browser-model-strategy current`, which skips `assertResolvedModelSelection` entirely and silently inherits whatever the account defaults to (it returned GPT-5.3 for me).

Tests added to `runOptions`, `cli/options`, `cli/browserConfig`. One existing assertion changed intentionally: `gpt-5.6-pro` no longer throws, since it now resolves; `GPT-5.6 Luna` / `GPT-5.6 Terra` still do.

Suite outcome unchanged: 8 pre-existing failures before and after (Azure / route-plan cases that also fail on a clean checkout in my environment).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Browser: run long local Pro consultations in a detached worker while the CLI remains attached to its session log, so unexpected foreground termination cannot stop answer capture; Ctrl-C still cancels the worker. Thanks @Rokurolize!
 - Browser: recognize ChatGPT's separate Pro effort control as the selected maximum effort for GPT-5.6 Sol when `--browser-thinking-time heavy` is requested. Thanks @Rokurolize!
+- Browser: accept `gpt-5.6-sol-pro` (also spelled `gpt-5.6-pro` / `GPT-5.6 Sol Pro`) as a selectable target so Sol's Pro effort can be requested directly, instead of being rejected as an unknown variant and then vetoed by the Pro-pill guard.
 
 ## 0.16.1 — 2026-07-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Browser: run long local Pro consultations in a detached worker while the CLI remains attached to its session log, so unexpected foreground termination cannot stop answer capture; Ctrl-C still cancels the worker. Thanks @Rokurolize!
 - Browser: recognize ChatGPT's separate Pro effort control as the selected maximum effort for GPT-5.6 Sol when `--browser-thinking-time heavy` is requested. Thanks @Rokurolize!
-- Browser: accept `gpt-5.6-sol-pro` (also spelled `gpt-5.6-pro` / `GPT-5.6 Sol Pro`) as a selectable target so Sol's Pro effort can be requested directly, instead of being rejected as an unknown variant and then vetoed by the Pro-pill guard.
+- Browser: accept `gpt-5.6-sol-pro` (also spelled `gpt-5.6-pro` / `GPT-5.6 Sol Pro`) as a selectable target so Sol's Pro effort can be requested directly, instead of being rejected as an unknown variant and then vetoed by the Pro-pill guard. The target is classified as a Pro model (detached worker), carries an explicit `apiModel` fallback, and verifies Sol identity from the picker rather than from the unlabelled Pro radio.
 
 ## 0.16.1 — 2026-07-23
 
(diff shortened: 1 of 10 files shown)


Learn2Solve:
Thanks — all three findings were real, and I verified each against the code before changing anything. Pushed in c6b3832.

**P1 "Preserve the Sol version when selecting Pro"** — correct, and worse than cosmetic: it reintroduced the exact silent-wrong-model failure this PR set out to remove. A bare `Pro` label left `desiredVersion` and `desiredModelVariant` null, so none of the guards fired and any active model's Pro effort would have verified as success.

The label now keeps its full identity (`GPT-5.6 Sol Pro`). Since the Pro radio's own text carries neither version nor variant, identity is read from the surrounding picker instead — `pickerShowsGpt56Sol()` requires the open menu (or the composer signal) to evidence Sol, and both the version/variant guards and `canTrustSelectedOption` defer to it for this tier.

**P1 "Route the browser-only alias to the real API model"** — correct. `resolveEffectiveModelId` returns `apiModel ?? model` and that id goes on the wire, so the config entry now sets `apiModel: "gpt-5.6-sol"` explicitly instead of leaning on the `model` field.

**P2 "Classify Sol Pro as a Pro model"** — correct. Added to `ProModelName` and `PRO_MODELS`; `isProModel()` is now true and `shouldDetachSession()` uses the detached worker. Confirmed live — the run reports `Detached session worker`.

One more spot the review did not name, but the same reasoning reaches it: `assertResolvedModelSelection` rejected any resolved label containing `pro` for a 5.6 target. That token *disqualifies* a plain Sol selection and is *required* for a Sol Pro one, so it is now compared against what was requested, and the error names the tier.

Re-verified live after the changes:

```
requested=GPT-5.6 Sol Pro; resolved=GPT-5.6 Sol Pro; verified=yes
Answer: GPT-5.6 Pro，Pro 推理档位。
```

Suite outcome unchanged: same 8 pre-existing failures as a clean checkout in this environment (Azure / route-plan cases).

steipete:
Closing per the maintainer review and the detailed analysis already posted above. The requested GPT-5.6 Sol/Pro behavior is now covered by the narrower, verified changes landed through #335 and #320/#348, including separate model and effort-pill handling. Keeping this parallel implementation would reintroduce overlapping picker logic without adding a distinct supported behavior.

Thank you for the contribution and investigation.

PR #345, opened 2026-07-30
[codex] hydrate ChatGPT conversations before recovery capture

pdurlej:
## Summary

Wait for a saved ChatGPT conversation to hydrate its prior turns before Oracle restarts response capture after a reload or reattach.

This prevents the browser runner from treating a correctly routed but still-empty ChatGPT SPA shell as the conversation. Recovery also no longer treats a shell-level Stop control as sufficient evidence when no user turn has rendered.

## Problem

ChatGPT can return from navigation with the durable `/c/<id>` URL and application shell visible before conversation turns hydrate. Oracle currently waits a fixed one second after stall/recheck navigation and begins capture immediately. The same race exists in live reattach and the new-Chrome recovery path.

In that state:

- `promptSubmitted` and model-selection evidence can both be valid;
- the URL can already identify the intended conversation;
- the turn list is still empty;
- a global Stop control may be visible;
- capture eventually times out or harvest returns UI placeholder text instead of an answer.

## Root cause

The resumed-conversation submission path already has `waitForResumedConversationHydration`, which waits for stable prior turns, confirms the composer has settled, and verifies the expected conversation id. Stall reload and reattach paths bypassed that gate and used either a fixed delay or no hydration wait.

## Fix

- Replace fixed post-navigation delays in local and remote delayed rechecks with the existing hydration gate.
- Gate the one-time stall reload before restarting assistant capture.
- Gate both live-target reattach and new-Chrome recovery before reading turn indexes.
- Require a hydrated user turn before a visible Stop control can mark recovery as ready.
- Add deterministic coverage for an empty shell that later hydrates, reattach ordering and fallback, and shell-only Stop controls.

No new selectors, endpoints, authentication behavior, or browser-profile access are introduced.

## Validation

- `pnpm exec vitest run tests/browser/index.test.ts tests/browser/reattach.test.ts tests/browser/recoverConversation.test.ts tests/browser/recoverConversationFlow.test.ts` - 94 passed
- `pnpm run check`
- `pnpm run build`
- `pnpm test` - 1,606 passed, 43 skipped

Exact-head neutral browser smoke on macOS:

- no files or private context;
- current Pro picker requested and independently verified;
- exact marker captured;
- transcript artifact persisted;
- browser slot released;
- 19.0 seconds end to end.

Exact head: `38f4acaf18b4d4faf785371ece301c467628f5f1`.

This complements #339: that change preserves the worker after foreground CLI loss, while this change prevents recovery capture from starting against a still-unhydrated conversation shell.


pdurlej:
We are consciously stopping further contributor-side work on this PR.

The integrated build works locally, the deterministic recovery coverage and CI are green, and the exact-head neutral browser smoke completed successfully. ClawSweeper found no actionable code or security defects. We also agree that the neutral smoke does not directly force the rare empty-shell -> hydration -> recovery-capture sequence, so we are not presenting it as proof of that specific live path.

We do not plan to spend additional browser runs trying to manufacture that condition. If a maintainer or contributor wants to take over, exercise the targeted reload/reattach path, or add redacted real-browser evidence, please do. The branch and tests are available for that work, and maintainers can also decide whether the deterministic evidence is sufficient.

steipete:
Maintainer repair pushed in `3242500d`.

History hydration and composer readiness are now separate contracts. Recovery waits for stable prior turns and the expected conversation, but it can start capturing an answer that is still running without waiting for the composer to re-enable. Normal resumed follow-up submission retains the stricter composer-ready gate.

Proof:

- 189 focused recovery/navigation tests passed (1 skipped)
- built-library recovery smoke hydrated 2 prior turns while a deliberately failing composer-ready callback was not invoked

Recommendation: **LAND** once refreshed checks complete.

PR #349, opened 2026-08-02
feat(api): add Atlas Cloud provider routing

binyangzhu000-sudo:
## Summary

- add an explicit `--provider atlas` route backed by `ATLASCLOUD_API_KEY`
- default Atlas Cloud runs to `https://api.atlascloud.ai/v1` while preserving arbitrary model IDs
- expose Atlas routing in provider diagnostics and auth recovery, with endpoint override support

## Validation

- `pnpm test` (1673 passed, 43 skipped)
- `pnpm run typecheck`
- `pnpm run format:check`
- focused `oxlint`
- `pnpm run docs:check`
- `pnpm run build`
- live CLI call through `deepseek-ai/deepseek-v4-pro` returned the expected response via the default Atlas endpoint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - API: add explicit GPT-5.6 reasoning mode and effort controls, including Pro mode, session persistence, long-run handling, and fail-closed route validation. Thanks @enki!
+- API: add explicit Atlas Cloud routing with `ATLASCLOUD_API_KEY`, a default compatible endpoint, and arbitrary model ID passthrough.
 
 ### Changed
 
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -611,9 +611,9 @@ program
   .addOption(
     new Option(
       "--provider <provider>",
-      "Choose API provider routing: auto, openai, or azure. Use openai to ignore Azure env/config.",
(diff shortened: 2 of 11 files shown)


steipete:
Thanks for the proposal and the focused routing tests. I’m closing this rather than adding a second provider-specific configuration contract.

Oracle already handles Atlas through the generic OpenAI-compatible path:

```text
OPENAI_API_KEY=<atlas-key> oracle --engine api --provider openai \
  --base-url https://api.atlascloud.ai/v1 \
  --model deepseek-ai/deepseek-v4-pro ...
```

I verified the built CLI’s route plan with that shape: it reports `provider: OpenAI-compatible`, `base: api.atlascloud.ai/v1`, the `OPENAI_API_KEY` source, and `ok`; a dry-run preserves the arbitrary model ID. The existing shared Chat Completions adapter therefore owns this transport without an Atlas-specific branch.

The proposed `--provider atlas` plus `ATLASCLOUD_*` variables would make Oracle maintain parallel key, endpoint, diagnostics, and recovery behavior for the same compatible-endpoint capability. That public-contract cost is not justified by a missing transport feature here. The PR also has no inspectable redacted artifact for its claimed live Atlas call, and contributor PRs should leave changelog ownership to release preparation.

If Atlas eventually needs behavior the generic adapter cannot express, a fresh proposal should start with that concrete incompatibility and live evidence rather than a convenience alias.

PR #359, opened 2026-08-06
fix(security): restrict session artifact file permissions to owner-only

bunlongheng:
## Problem

Oracle sends a run's full context (the prompt, every attached file's contents, and the model's streamed response) and persists all of it under `~/.oracle/sessions/<id>/`. Those artifacts were created with the process umask default, leaving them readable by every local account:

| Path | Before | Contains |
|------|--------|----------|
| `sessions/`, `sessions/<id>/`, `<id>/models/` | `0755` | (traversable) |
| `<id>/output.log`, `models/*.log` | `0644` | full prompt + attached file contents + model response |
| `<id>/models/*.json` | `0644` | usage, response ids |
| `<id>/meta.json` | `0600` | already locked down |

On a shared or multi-user host, any other local user can read another user's oracle history - which routinely includes proprietary source code and can include secrets the user attached to the consult. This is a local information disclosure (CWE-732 *Incorrect Permission Assignment for Critical Resource* / CWE-200).

The inconsistency is the tell: `meta.json` is already written with `mode: 0o600` (`writeSessionMetadataFile`), and the bridge config/token files use `0o600`/`0o700` (`src/cli/bridge/host.ts`, `src/bridge/userConfigFile.ts`) - but the sibling files that hold the **most** sensitive content were left at the umask default.

### Reproduction (before)
```
$ oracle -p "review this" --file ./src   # sends code to GPT
$ ls -ld ~/.oracle/sessions/<id> ~/.oracle/sessions/<id>/output.log
drwxr-xr-x  ... ~/.oracle/sessions/<id>          # 0755 - traversable by anyone
-rw-r--r--  ... ~/.oracle/sessions/<id>/output.log  # 0644 - world-readable transcript
```

## Fix

Align the session artifacts with the posture already used for `meta.json` and the bridge files:

- session/model **directories** created `0o700`
- `output.log`, per-model `.log` and `.json` files written `0o600`
- the appended log stream in `createSessionLogWriter` opened with `mode: 0o600`

Two named constants (`SESSION_DIR_MODE`, `SESSION_FILE_MODE`) keep it consistent and self-documenting. No behavior change on single-user machines; on Windows the mode bits are effectively ignored (test skips there).

Note: file mode is applied at creation, so this hardens newly created sessions. Existing session directories keep their prior permissions unless recreated.

## Test

Added `session directory and log artifacts are owner-only` in `tests/sessionManager.test.ts`, asserting `sessions/`, the session dir, and `models/` are `0700` and `output.log` / `models/*.json` / `models/*.log` are `0600`.

- Before the fix the test fails: `expected 493 to be 448` (0o755 vs 0o700).
- After the fix the full session suite passes (`sessionManager`, `sessionStore`, `sessionStore.prune`, plus `mcp/sessions`, `cli/sessionCommand`, `mcp/consult`).
- `tsc --noEmit`, `oxlint`, and `oxfmt --check` all clean on the changed files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Fixed
 
+- Security: restrict existing and newly created session transcripts, model metadata, and browser artifacts to the current user, without following symlinks during upgrade hardening. Thanks @bunlongheng!
 - Browser: select reasoning effort in ChatGPT's unified Intelligence picker, where the tiers moved behind an `Advanced` → `Effort` submenu next to a power slider. `--browser-thinking-time` matched nothing in that layout and silently kept whatever effort the tab already had, which also left the Pro tier unreachable. Submenu openers are no longer treated as selectable tiers, so a `Model` row reading `GPT-5.6 Pro` cannot satisfy a Pro request, and a composer pill showing the Pro effort is no longer mistaken for a Pro model.
 
 ## 0.17.1 — 2026-08-02
--- a/src/sessionManager.ts
+++ b/src/sessionManager.ts
@@ -359,9 +359,13 @@ const DEFAULT_SLUG = "session";
 const MAX_SLUG_WORDS = 5;
 const MIN_CUSTOM_SLUG_WORDS = 3;
 const MAX_SLUG_WORD_LENGTH = 10;
+// Session artifacts (prompt, attached file contents, model responses) are sensitive.
+// Keep them owner-only, matching the meta.json / bridge-config posture (0o600/0o700).
+const SESSION_DIR_MODE = 0o700;
+const SESSION_FILE_MODE = 0o600;
 
 async function ensureDir(dirPath: string): Promise<void> {
-  await fs.mkdir(dirPath, { recursive: true });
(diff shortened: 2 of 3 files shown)


steipete:
Landed as `b75a720e9c4cece42c33962ce61ef78fa979ef7e`.

Verification:

- Packed-package black-box proof ran the real `oracle status` against a synthetic pre-upgrade tree: session directories changed from `0755` to `0700`, regular logs/model metadata/nested browser artifacts changed from `0644` to `0600`, and file/directory symlinks plus their external targets remained untouched.
- `pnpm vitest run tests/sessionManager.test.ts tests/browser/artifacts.test.ts` — 33 passed.
- Full local gate passed: format, typecheck/lint, docs check/site, 1,691 tests (43 skipped), build, and packed-CLI smoke.
- The first full-suite pass had one unrelated five-second timeout in the delayed ChatGPT image test; that exact test passed twice in isolation (59 ms and 21 ms), and the full retry passed.
- GitHub Actions run `31291574792`, attempt 2, passed on the exact landed head after fork-workflow approval; GitGuardian passed.
- Full-branch autoreview reported no accepted/actionable findings.

Issue #336, opened 2026-07-24
GPT 5.6 Sol Pro support?


rosstex:
Awesome repo! I saw #305 seems to add ChatGPT 5.6 Sol Pro support, is there a reason the docs/repo haven't been updated to show that 5.6 Sol Pro as the primary ChatGPT model?

steipete:
Yes — support is implemented and documented. Browser mode: `--model gpt-5.6-sol --browser-thinking-time pro`; API mode: `--model gpt-5.6-sol --reasoning-mode pro` (landed via #335, #361, #377). Support is separate from the unconfigured default, which stays tracked in #375, and the newly reported direct-slider regression is tracked in #422 (fixed by #424, merged). Closing this support question — it does not decide the default policy.

PR #354, opened 2026-08-03
fix(browser): keep extra-high distinct from Pro thinking time

ozeron:
## Summary

- Stop normalizing `extra-high` / `xhigh` / `extrahigh` to `heavy`
- Treat `extra-high` as Extra High (non-Pro) and `heavy` as explicit Pro on GPT-5.6 Sol
- Update skill + CLI/config docs so agents no longer request `heavy` for Extra High

Fixes #353

## Proof

- `pnpm vitest run tests/cli/options.test.ts tests/browser/thinkingTime.test.ts tests/mcp/consult.test.ts tests/mcp/chatgptImage.test.ts tests/cli/browserDefaults.test.ts` — 127 passed
- `pnpm run build`
- Live browser smoke (built CLI, GPT-5.6 Sol, `--browser-thinking-time extra-high`, copy-profile):
  - session `pr354-extra-high-proof`
  - `meta.json` → `browser.config.thinkingTime: "extra-high"`
  - log → `[browser] Thinking time: Extra High` (not Pro)
  - model evidence → `resolvedLabel=GPT-5.6 Sol`; answer `ok`
- Docs/skill updated: Sol max effort is `extra-high`; `heavy` reserved for Pro

@clawsweeper re-review

PR #361, opened 2026-08-08
fix(browser): drive ChatGPT's Advanced → Effort submenu for reasoning effort

mkubenka:
## Problem

ChatGPT's unified Intelligence picker moved the effort tiers. The simple view is now a 5-position power slider; the tiers live behind an `Advanced` toggle that reveals `Model` and `Effort` submenu openers.

Oracle only scanned the top-level menu, so on this layout **every** `--browser-thinking-time` value matched nothing and the run continued at whatever effort the tab already had:

```
$ oracle --model gpt-5.6-sol --browser-thinking-time extra-high -p "HI"
[browser] Thinking time: option not found (requested Extra-high);
          keeping the effort already selected in ChatGPT.
```

`strictProEffort` did not catch this, because it only arms when the target *model* kind is Pro — impossible here, since Pro is now a tier rather than a model row. So it degraded with a warning instead of failing.

The Pro tier was unreachable by any flag. `--model gpt-5-pro` / `gpt-5.5-pro` (the default) fail earlier still, because there is no `Pro` row in the model switcher:

```
ERROR: Unable to find model option matching "Pro" in the model switcher.
Available: Advanced, ModelGPT-5.6 Sol, EffortHigh.
```

That last one means the **default invocation** is broken for anyone on this picker.

## Observed DOM (2026-08-07)

One `role="menu"` containing:

| node | notes |
|---|---|
| `div[data-testid="composer-model-picker-slider-simple-view"]` | "High, 3 of 5. Use Left and Right arrow keys to adjust power." |
| `div[role=menuitem][aria-label="Power"]` | the slider |
| `div[role=menuitem][aria-label="Show advanced options"][aria-expanded=false]` | "Advanced" |
| `div[data-testid="composer-model-picker-slider-advanced-view"]` | "ModelGPT-5.6 SolEffortHigh" |
| `div[role=menuitem][aria-haspopup=menu][data-state=closed]` | "ModelGPT-5.6 Sol" |
| `div[role=menuitem][aria-haspopup=menu][data-state=closed]` | "EffortHigh" |

`trailingCount: 0` — no `[data-model-picker-thinking-effort-action]` nodes exist. Activating `Effort` opens a submenu of `Instant / Medium / High / Extra High / Pro`.

Note that labels concatenate without separators (`EffortHigh`).

## Change

- **Descend into the submenu** when the flat scan finds no tier: expand `Advanced` if collapsed, activate `Effort`, resolve via `aria-controls`, match with the existing level tokens. Older layouts are untouched — the descent only runs after the flat scan returns null.
- **Identify the Effort opener positively, by name.** The sibling Model opener has an identical `role`/`aria-haspopup` shape and can read `ModelGPT-5.6 Pro`, so guessing from tier words picks the wrong control and then clicks model rows as if they were efforts. An unrecognized language yields no opener and the caller fails rather than guessing.
- **New explicit `pro` level**, not an overload of `heavy`, and the fail-closed guard now arms for it — Pro is expensive and rate-limited, so an unconfirmed selection aborts instead of quietly submitting at a cheaper tier.
- **`pro` stays out of `LEVEL_TOKENS`**, which classifies whether a control or menu is an effort owner. "Pro" is also a model name; counting it there makes model pills and model menus look like effort controls. `TARGET_LEVEL_TOKENS` adds it for target matching only.
- **Submenu openers are no longer selectable tiers**, and a composer pill showing the Pro *effort* is no longer read as a Pro *model*. `Pro Extended` still names a legacy Pro model plus its effort and keeps the original inference — otherwise a Pro Extended user asking for `extended` gets moved down to High.

## Verification

Live, same account, same commands as above:

```
[browser] Thinking time: Extra High                       (29.2s)
[browser] Thinking time: Pro                              (41.0s)
[browser] Thinking time: Pro (already selected)           (34.3s)
```

Both the `switched` and `already-selected` branches exercised, plus a 13-minute Pro run on a real workload.

`1688 passed`, typecheck and lint clean. 12 new tests, including: the submenu is unreachable until its opener is activated (so a descent that never opens it cannot pass); a Model opener labelled `Pro`; an unrecognized language; a hover-reactive decoy tier row; and all seven failure statuses for explicit Pro.

## Notes for review

- Relationship to #342: same root problem, but that PR's DOM (2026-07-27) shows a flat `menuitemradio` list, which the picker has since replaced. It also makes `gpt-5.6-sol-pro` a *model* target mapping to the `Pro` label — on this layout there is no `Pro` model row to match. This PR treats Pro as an effort tier instead.
- Not addressed here: `heavy` still accepts an already-Pro pill (#340), so Heavy can run Pro when the tab is already there. Left alone as deliberate existing behaviour, though it does conflict with keeping `heavy` and `pro` separate.
- The default model target is still `Pro` (`src/browser/constants.ts`), so a bare `oracle` run keeps failing on this picker with the error above. Changing that default looks like a product call rather than a bug fix — see #336 — so it is deliberately left out of this PR.


mkubenka:
Thanks — two of the three findings were real and are fixed in fdb5c8f5. Pushing back on the third.

### 1. Advanced treated as an Extended effort row — fixed, and it was worse than described

Confirmed: `erweitert` is in the `extended` tier tokens *and* is the German Advanced label, and the Advanced toggle carries no `aria-haspopup`, so the opener filter did not exclude it.

Reproduced both failure modes against a German fixture:

```
pill on High,    requesting extended -> {"status":"switched","label":"Erweitert"}   effortOpener clicks: 0
pill on Instant, requesting extended -> {"status":"selection-unverified"}           effortOpener clicks: 0
```

The first is the more serious one: it reports **success** having only clicked the disclosure toggle, because `currentEffortPillMatchesTarget` then sees a pill that already reads "High".

Fixed structurally rather than by extending a word list, since any label-based rule has the same collision problem in the next locale. A real tier row carries a checked state; a container carries expansion state — either a submenu opener (`aria-haspopup=menu`) or a disclosure toggle (`aria-expanded` with no `aria-checked`). Neither is selectable, so neither can satisfy a tier request. Same fixture after the fix:

```
pill on High,    requesting extended -> {"status":"already-selected","label":"High"}  effortOpener clicks: 1
pill on Instant, requesting extended -> {"status":"switched","label":"High"}          effortOpener clicks: 1
```

Added the localized regression test you asked for. It fails if the disclosure-toggle exclusion is reverted (verified by mutation, not just by it being green).

### 2. Document the new `pro` value — fixed

Updated `docs/cli-reference.md`, `docs/browser-mode.md`, `docs/configuration.md`, and `docs/mcp.md`, including the fail-closed behaviour and the fact that an unrecognized `Effort` label makes Oracle decline rather than guess. `pnpm docs:check` passes.

### 3. Release-owned changelog — respectfully disagree, leaving as is

This appears to contradict the repository's own instructions and its merge history:

- `AGENTS.md:29`: *"After finishing a feature, ask whether it matters to end users; if yes, update the changelog. Read the top ~100 lines first and group related edits into one entry instead of scattering multiple bullets."* The entry here is deliberately grouped into one `Added` line and one `Fixed` line for exactly that reason.
- Every recent merged PR I checked touches `CHANGELOG.md`: #332, #343, #345, #354, #357.

Happy to drop it if the convention has changed — but as documented, omitting it would be the deviation.

---

`1689 passed`, typecheck, lint and `docs:check` clean. The three CI builds and the CDP proof job were green on the previous head and this change is additive to them.

steipete:
Landed as `573b6c7d87e8d2ea9556f2a9596585975017e1de`.

Verification:

- `pnpm run check && pnpm run docs:check && pnpm run docs:site && pnpm run test && pnpm run build && pnpm run test:packed-cli` — passed; 1,689 tests passed and 43 skipped.
- Packed-package black-box smoke accepted `--browser-thinking-time pro` and rejected an invalid level with the updated allowed-values message.
- Signed-in live ChatGPT smoke selected `GPT-5.6 Sol` and reported `[browser] Thinking time: Pro` on the current unified picker.
- GitHub Actions run `31291061485` passed Ubuntu, macOS, Windows, and the Linux CDP disconnect proof; GitGuardian passed.
- Full-branch autoreview reported no accepted/actionable findings.

Caveat: my fresh live smoke hit an unrelated downstream prompt-commit timeout after Pro was selected; the composer remained populated with no picker menu open and an enabled send button. The PR author's successful switched/already-selected Pro runs remain the complete submission proof.

PR #366, opened 2026-08-12
feat(browser): support Grok web sessions

daxiondi:
## Summary

- add a dedicated Grok DOM browser executor for attached Chrome sessions at `grok.com`
- support text prompts and same-conversation browser follow-ups while failing closed for unsupported attachments and remote-host runs
- preserve existing xAI API auto-routing unless browser mode is explicitly requested
- detect Cloudflare verification and anonymous sign-up walls with actionable errors

User impact: xAI subscribers can use their existing Grok web login with Oracle browser mode instead of requiring an API key. The executor always opens an isolated Grok tab, so unrelated active browser tabs are not modified.

## Test Plan

- `pnpm format:check`
- `pnpm lint`
- `pnpm test` (1705 passed, 43 skipped)
- `pnpm build`
- real Chrome/CDP smoke against `https://grok.com` via `--remote-chrome 127.0.0.1:53388`: verified isolated-tab creation, current Tiptap prompt entry, submission, and immediate anonymous sign-up-wall detection

The available Chrome profile was not signed into Grok, so authenticated assistant-response extraction could not be exercised live in this environment. Unit coverage exercises streaming-to-stable response extraction and follow-up behavior.

## Risk / Rollback

Risk is isolated to explicit `--engine browser --model grok` runs. Grok DOM changes could require selector updates; explicit login/challenge/timeout errors avoid silently returning incorrect content. Attachments and `--remote-host` fail closed until those paths have dedicated implementations.

Rollback: revert commit `26afeb2b` or continue using `--engine api --model grok`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Browser: add Grok web automation through an attached `grok.com` Chrome session, including text prompts and same-conversation follow-ups.
+- Browser: add Grok web automation through an attached `grok.com` Chrome session, including text prompts, file uploads, same-conversation follow-ups, and final-response extraction that excludes Grok's visible thinking process.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.17.3 — Unreleased
 
+### Added
+
+- Browser: add Grok web automation through an attached `grok.com` Chrome session, including text prompts and same-conversation follow-ups.
+
 ### Fixed
 
 - Browser: recognize the Japanese `詳細設定` → `推論レベル` controls in ChatGPT's unified Intelligence picker, allowing explicit Pro effort selection without weakening the fail-closed guard for unknown languages.
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -469,7 +469,7 @@ program
   .addOption(
     new Option(
       "-e, --engine <mode>",
-      "Execution engine (api | browser). Browser engine: GPT models automate ChatGPT; Gemini models use a cookie-based client for gemini.google.com. If omitted, oracle picks api when OPENAI_API_KEY is set, otherwise browser.",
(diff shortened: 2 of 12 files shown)


daxiondi:
Authenticated Grok web smoke completed after selector fix (commit `d1932303`).

Live command used the signed-in GeekEZ Chrome session on `127.0.0.1:53388`. The run opened an isolated Grok tab, submitted the prompt, ignored `Thought for 3s`, waited for `.response-content-markdown`, and returned exactly `GROK_WEB_AUTH_OK_2` in 10.5s.

Additional verification: `pnpm format:check`, targeted Grok executor tests (4/4), `pnpm typecheck`, and `pnpm build` passed.

daxiondi:
Fixed two validated Grok browser regressions in `b38ab787`:

1. **Real file upload**: Grok attachments now use CDP `DOM.setFileInputFiles`, wait for the filename/chip, and submit only after upload readiness. Follow-ups do not re-upload the initial files.
2. **Final answer vs thinking**: response extraction now requires non-empty `.response-content-markdown`/`.markdown` and a finished generation state, rather than returning `Thought for Ns` or agent-process text.

Live evidence:
- Uploaded `grok_query.md` with `--browser-attachments always`; the Grok conversation contained the filename and produced the complete four-question Chinese red-team review after **2m21s**, while visible thinking lasted **1m31s** and was excluded from Oracle output.
- Updated the globally installed Oracle build and ran a falsifiable attachment smoke asking only for line 10. Grok returned exactly `60-140` with `files=1`.
- Full validation: format, lint, typecheck, build, and tests passed (`1705 passed`, `43 skipped`).

steipete:
Thanks @daxiondi for the working implementation — the effort here is appreciated and the code demonstrated the flow end-to-end.

After weighing it, we're declining Grok web-session support: oracle's browser engine currently targets one provider surface, and every additional web target multiplies the selector-breakage surface we have to chase on every UI redesign (the recent ChatGPT unified-picker migration is a fresh reminder of that cost). We'd rather keep the browser engine reliable for its primary target than broaden it.

If Grok ships a stable API surface, that would be the integration path we'd take instead. Closing.

PR #369, opened 2026-08-13
browser: bound Pro answer-now placeholder detection

enieuwy:
## Problem

Oracle needs to ignore ChatGPT's short `Pro thinking` / `Answer now` placeholder while generation is still running. Capturing that UI-only value would return an incomplete response.

The current guard is broader than that intent. `isAnswerNowPlaceholderText()` runs against the extracted text of the whole assistant turn and returns `true` whenever that value contains both `answer now` and either `pro thinking` or `chatgpt said` anywhere:

```ts
return (
  text.includes("answer now") &&
  (text.includes("pro thinking") || text.includes("chatgpt said"))
);
```

Consequently, substantive assistant content is discarded whenever those strings are present in the same extracted value. Every extraction layer applies this predicate, so Oracle keeps waiting instead of passing the content to its normal completion gate.

This is a false-positive repair, not removal of the incomplete-response guard.

## Change

- Continue treating known short, UI-only `Pro thinking` / `Answer now` combinations as placeholders.
- Treat a value as a placeholder only when its whole normalized text matches that bounded UI grammar.
- Preserve substantive assistant text that merely contains those strings.
- Use the same closure-free predicate in Node-side parsing and both injected page expressions, so the extraction paths cannot drift.

The 60-character bound covers the longest known UI-only composite (57 characters). Unknown substantive text falls through to Oracle's existing completion checks; this PR does not bypass the stop/thinking/action-bar terminal gate.

## Behavior proof

A real after-fix ChatGPT Pro browser run intentionally requested a two-paragraph answer containing both trigger strings. This exercises the deterministic false-positive path in the upstream predicate: current upstream would classify the returned value as a placeholder solely because both substrings are present.

Command (browser path redacted):

```bash
oracle --engine browser \
  --browser-chrome-path /path/to/chromium \
  --browser-headless \
  --browser-timeout 300s \
  --model gpt-5.6-sol \
  --browser-thinking-time pro \
  --prompt 'Write two short paragraphs about why TCP retries. Include the words Pro thinking and Answer now somewhere in the prose. End with exactly: HARVEST_PLACEHOLDER_SMOKE_OK'
```

Observed output:

```text
[browser] Thinking time: Pro (already selected)
[browser] ChatGPT thinking - 30s elapsed; status=response streaming; source=inline
[browser] Archived ChatGPT conversation after saving local artifacts.
[browser] Model selection evidence: requestedKey=gpt-5.6-sol; target=GPT-5.6 Sol; resolvedLabel=GPT-5.6 Sol; status=already-selected; strategy=select; verified=yes; source=chatgpt-model-picker

Answer:
TCP retries because packets or acknowledgements can be lost, corrupted, or delayed in transit. When the sender does not receive confirmation within the expected time, it retransmits the missing data; Pro thinking about sequence numbers and checksums explains how TCP avoids accepting duplicates.

TCP also uses duplicate acknowledgements and adaptive timers to detect loss without retrying too aggressively. Answer now: these mechanisms provide reliable, ordered delivery over an unreliable network. HARVEST_PLACEHOLDER_SMOKE_OK

57.2s · GPT-5.6 Sol[browser] · ↑42 ↓132 ↻0 Δ174
```

Saved session evidence:

```json
{
  "status": "completed",
  "mode": "browser",
  "browser": {
    "runtime": { "promptSubmitted": true },
    "modelSelection": {
      "requestedModel": "GPT-5.6 Sol",
      "resolvedLabel": "GPT-5.6 Sol",
      "status": "already-selected",
      "verified": true,
      "source": "chatgpt-model-picker"
    },
    "warnings": []
  },
  "elapsedMs": 57154,
  "usage": {
    "inputTokens": 42,
    "outputTokens": 132,
    "reasoningTokens": 0,
    "totalTokens": 174
  }
}
```

This proves the repaired path returns substantive Pro output containing both trigger strings while retaining normal response-streaming and terminal-completion handling. It does **not** claim to be a preserved raw-DOM capture of mounted skip-ahead controls; the source defect does not depend on why those strings enter the extracted turn.

## Automated verification

- `pnpm vitest run tests/browser/assistantResponseStatus.test.ts` (93 tests)
- `pnpm run check`
- `pnpm run docs:check`
- GitHub CI: Ubuntu, macOS, Windows, CDP disconnect proof, and GitGuardian all pass

Regression coverage includes:

- long substantive answers containing both trigger strings;
- genuine short UI-only placeholder combinations;
- Node-side classification;
- execution of the injected predicate in snapshot and response-observer page contexts.

## Risk

If ChatGPT adds a new short UI-only label combination outside the known grammar, it could pass into the existing completion checks rather than being filtered immediately. The shared predicate and focused fixture table make additions explicit; the PR does not alter Oracle's independent positive terminal-completion gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Browser: recognize the Japanese `詳細設定` → `推論レベル` controls in ChatGPT's unified Intelligence picker, allowing explicit Pro effort selection without weakening the fail-closed guard for unknown languages.
+- Browser: treat the skip-ahead control labels as placeholder signals only when the entire turn text is short chrome text, so substantial assistant answers that mention those labels are no longer discarded.
 
 ## 0.17.2 — 2026-08-10
 
--- a/src/browser/actions/assistantResponse.ts
+++ b/src/browser/actions/assistantResponse.ts
@@ -138,21 +138,70 @@ function matchesThinkingStatusLabel(trimmed: string): boolean {
   return trimmed.startsWith("pro thinking") && trimmed.length <= 40;
 }
 
-export function isAnswerNowPlaceholderText(value: string): boolean {
-  const text = value.toLowerCase().replace(/\s+/g, " ").trim();
-  if (!text) return false;
-  // Learned: "Pro thinking" shows a placeholder turn that contains "Answer now".
-  // That is not the final answer and must be ignored in browser automation.
-  if (text === "chatgpt said:" || text === "chatgpt said") return true;
-  if (
-    text.includes("file upload request") &&
-    (text.includes("pro thinking") || text.includes("chatgpt said"))
-  ) {
-    return true;
+// Single source of truth for the "Answer now" placeholder test.
+//
+// This function is ALSO injected verbatim into page expressions via
+// buildAnswerNowPlaceholderPredicateJs (it is stringified with
+// Function.prototype.toString), so it MUST stay closure-free: no imports, no module
+// constants, no helpers. Anything it references must be declared inside its own body
+// or it will throw a ReferenceError inside the renderer.
+//
+// Learned the hard way: ChatGPT's Pro UI keeps the "Answer now" skip-ahead control
+// mounted for the whole reasoning phase, so a bare `text.includes('answer now')` test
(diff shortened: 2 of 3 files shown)


enieuwy:
@clawsweeper re-review

PR #370, opened 2026-08-13
browser: prove composer commit from ProseMirror state

enieuwy:
## Summary

- read ChatGPT composer state from `pmViewDesc.node` (`textBetween`) with a React-props fallback
- compare a Markdown-invariant projection so heading/list/emphasis syntax consumed into node types cannot fail a healthy prompt
- call `onPromptSubmitted` only after `verifyPromptCommitted` succeeds

ChatGPT's ProseMirror composer often exposes no `pmViewDesc.view` and no useful `__reactProps$` on the editor node. A DOM `innerText` check is not framework state: Markdown input rules consume `##`, lists, and emphasis into nodes and marks, so literal equality is unachievable for Markdown-bearing prompts.

This change proves the prompt the composer will submit, then marks submission only after the turn actually commits. It does not add large-prompt insertion machinery or extra Chrome flags.

## Verification

- `pnpm vitest run tests/browser/promptComposer.test.ts` (12 tests)
- `pnpm run check`
- `pnpm run docs:check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,308 +1,958 @@
-[/private/tmp/oracle-composer/CHANGELOG.md#B261]
-1:# Changelog
-2:
-3:## 0.17.3 — Unreleased
-4:
-5:### Fixed
-6:
-7:- Browser: recognize the Japanese `詳細設定` → `推論レベル` controls in ChatGPT's unified Intelligence picker, allowing explicit Pro effort selection without weakening the fail-closed guard for unknown languages.
-8:- Browser: read the prompt composer through the editor's framework state when available so the current ChatGPT composer is correctly recognised as filled before sending, with a fallback for other builds.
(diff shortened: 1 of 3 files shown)


enieuwy:
Closing this revision rather than weakening Oracle's submission-safety contract.

Review found two independent blockers:

- `projectForComparison()` is non-injective: lowercasing and stripping Markdown markers/newline runs lets case changes and structural loss compare equal (for example, `## Heading` vs `Heading`, or two paragraph breaks vs one).
- Moving `onPromptSubmitted` after `verifyPromptCommitted()` weakens the deliberate dispatch boundary. A send can succeed while commit probing times out; leaving `promptSubmitted=false` then permits recovery to resend and duplicate the request.

A safe replacement needs an authoritative, exact node/mark representation while preserving prompt-submitted persistence at the dispatch boundary. This PR does not provide that contract, so I am withdrawing it rather than adding more lossy parsing.

PR #371, opened 2026-08-13
browser: sync cookies during authenticated session recovery

enieuwy:
## Problem

Oracle uses cookie synchronization to seed authentication before the first ChatGPT navigation. Initial launch treats manual-login profiles as pre-signed and synchronizes them only when both `cookieSync` and `manualLoginCookieSync` are enabled.

Session recovery used `cookieSync && !manualLogin`. A persistent manual-login profile configured with that explicit opt-in could therefore receive cookies during initial launch but not when Oracle relaunched Chrome to reopen the saved conversation.

The documented `browser.manualLoginCookieSync` user setting was also dropped while converting user configuration into a one-shot browser session, so that CLI path could not activate this explicit opt-in.

## Change

- Apply the same manual-login cookie-sync opt-in during new-Chrome session recovery as during initial launch.
- Continue skipping cookie sync for ordinary manual-login profiles, which reuse their persistent profile cookies.
- Apply configured Chrome/inline cookies when `manualLoginCookieSync` explicitly opts a manual-login profile into synchronization.
- Forward `browser.manualLoginCookieSync` through CLI defaults and browser-session construction.
- Attempt to close the recovery CDP client and, unless `keepBrowser` is enabled, terminate the launched Chrome before propagating a cookie-sync failure.

## Verification

The cookie-sync-focused tests set `cookieSync` explicitly, so their expectations are independent of platform defaults. The recovery-path regression calls public `resumeBrowserSession()` without a live Chrome endpoint, lets it select the default new-Chrome recovery path, and injects launch/connect/cookie-sync dependencies. It proves that explicit manual-login synchronization invokes `syncCookies`; the failure case also proves that the CDP client is closed and the launched Chrome is killed under the default `keepBrowser: false` policy before the error is returned.

- `pnpm vitest run tests/cli/browserDefaults.test.ts tests/cli/browserConfig.test.ts tests/browser/reattach.test.ts` (97 tests)
- `pnpm run check`
- `pnpm run build`
- `pnpm run docs:check`

## Redacted real-browser recovery proof

Revision `14810ecafb27` first created an unarchived conversation through the configured manual-login cookie-sync path. A separate one-shot harness then called public `resumeBrowserSession()` with no live Chrome endpoint and a newly created target profile containing only an empty `Default/` directory. It supplied the saved conversation URL and an authenticated local Chrome profile as redacted environment values.

The harness delegated to Oracle's production `launchChrome()` and `syncCookies()` implementations, wrapping them only to record the launched PID and applied-cookie count. Chrome connection, cookie application through CDP, authentication checks, saved-conversation navigation, DOM response capture, and cleanup all used the production recovery path.

```text
proof_commit=14810ecafb27
source_worktree_clean=true
new_chrome_launched=true
target_profile_started_empty=true
manual_login_cookie_sync=enabled
cookies_applied=3
authenticated_conversation_reopened=true
answer=PR371_RECOVERY_PROOF_OK
launched_chrome_stopped=true
proof_status=passed
```

The empty target profile rules out authentication inherited from the recovery profile. Cookie values, account data, local paths, browser identifiers, and the private conversation URL are omitted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Browser: recognize the Japanese `詳細設定` → `推論レベル` controls in ChatGPT's unified Intelligence picker, allowing explicit Pro effort selection without weakening the fail-closed guard for unknown languages.
+- Browser: align reattach recovery cookie sync with the launch path for manual-login profiles with explicit cookie sync, so recovery can reopen an authenticated conversation with supplied cookies instead of skipping sync.
 
 ## 0.17.2 — 2026-08-10
 
--- a/src/browser/index.ts
+++ b/src/browser/index.ts
@@ -109,6 +109,7 @@ import {
   resolveManualLoginWaitMs,
 } from "./manualLoginProfile.js";
 import { describeBrowserControlPlan, formatBrowserControlPlan } from "./controlPlan.js";
+import { shouldSyncBrowserCookies } from "./policies.js";
 import {
   createConversationUrlMonitor,
   type ConversationUrlMonitor,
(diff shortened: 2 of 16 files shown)


enieuwy:
@clawsweeper re-review

enieuwy:
@clawsweeper re-review

PR #302, opened 2026-07-10
fix(browser): restore prompt submission when the Chrome window is hidden (#298)

LeoLin990405:
Fixes #298 and #312.

## Root cause

PR #282 moved prompt submission to trusted CDP mouse input. On macOS, `--browser-hide-window` then hid the Chrome process with Cmd-H. Hidden Chrome stopped compositing the page, swallowed the trusted click without an error, and retained the prompt as a draft. A later visible run could submit accumulated drafts together.

## Final design

- Launch hidden macOS Chrome off-screen with `--window-position=-32000,-32000`, keeping it normally rendered from startup.
- Reposition reused Chrome windows through CDP without minimizing or hiding the process.
- Keep focus emulation as a soft-failing local/remote optimization.
- Remove the proposed post-click synthetic fallback. It could mistake historical assistant content for a successful new send, or race a delayed trusted click and submit twice.
- Preserve the existing baseline-aware prompt-commit verifier as the only commit proof.
- Apply the same off-screen behavior to normal browser runs, Project Sources, and recovered sessions.

Contributor commit and authorship are preserved. The maintainer hardening commit replaces only the unsafe fallback strategy.

## Tests

- `pnpm vitest run tests/browser/chromeLifecycle.test.ts tests/browser/controlPlan.test.ts tests/browser/promptComposer.test.ts tests/browser/index.test.ts` — 84 passed.
- `pnpm check` — passed.
- `pnpm test` — 1,436 passed, 43 skipped.
- Final AutoReview — no actionable findings.

Regression coverage includes off-screen launch/repositioning, historical assistant content not proving a new commit, and a delayed trusted click remaining a single send attempt.

## Live macOS Pro proof

Exact branch tree, signed-in ChatGPT Pro, `--copy-profile`, `--browser-hide-window`, model selection verified as Pro:

1. `pr302-hidden-followup-live`: initial marker and same-conversation follow-up both returned exactly once in 22.3s.
2. `pr302-hidden-second-live`: a fresh consecutive hidden-window run returned only its marker in 10.8s.

No prompt remained in the composer, no prior marker leaked into the next run, and no duplicate/combined send occurred.

Exact head: `05b8897f4cf21beb4d9bbdf4327ae31b7e5fa919`.

PR #380, opened 2026-08-14
fix(browser): restore visible macOS window bounds

IndelibleVivi:
## Summary

- record a persistent macOS profile's last Chrome placement before a fresh hidden launch applies the off-screen launch flag
- restore a locally launched window only when Oracle's profile-local saved-state marker proves ownership
- preserve exact normal bounds and maximized/fullscreen state, then remove the marker after a successful visible restore
- leave every unmarked window untouched, including valid negative-coordinate placements on another display
- apply the policy to normal browser runs, Project Sources, and new-Chrome reattach while leaving headless, attach-running, and remote Chrome untouched

## Context

#302 replaced app hiding with off-screen window positioning so macOS Chrome continues rendering trusted CDP input.

A persistent Oracle Chrome window can remain at those off-screen bounds when a later run uses the default visible policy. Current main omits the off-screen step when `hideWindow` is false, but it does not restore a window that Oracle previously moved.

For an already-running local Chrome, Oracle records the live CDP bounds immediately before moving the window. For a fresh hidden launch, Oracle reads the selected profile's persisted `browser.window_placement` and writes the marker before constructing and applying `--window-position=-32000,-32000`. A later visible run restores only from that marker. Geometry alone is never treated as ownership, so markerless windows are not queried or repositioned.

## Verification

- `pnpm exec vitest run tests/browser/chromeLifecycle.test.ts tests/browser/index.test.ts tests/browser/reattach.test.ts tests/browser/projectSources.test.ts`
  - 108 passed
- `pnpm run check`
  - passed
- `pnpm run build`
  - passed
- `pnpm test`
  - 1,764 passed, 43 skipped

The macOS launch-order regression test invokes the real `launchChrome` function with a mocked launcher and confirms both facts at the launch boundary: the exact persisted placement marker already exists, and the outgoing flags contain `--window-position=-32000,-32000`.

A built-code Chrome for Testing probe used a disposable unsigned profile and no ChatGPT/account data. The redacted lifecycle trace was:

```text
persisted profile placement  (240, 120, 1100 x 760)
pre-launch Oracle marker     (240, 120, 1100 x 760)
post-hide macOS bounds       (-1240, 29, 1280 x 720)
visible restore              (240, 120, 1100 x 760)
marker after restore         absent
unmarked negative placement  (-320, 120, 1100 x 760) -> unchanged
```

The disposable browser process and profile were removed after the probe.


IndelibleVivi:
Addressed the visible-window compatibility finding in `7840c0cb`.

- Hidden local runs now save the exact pre-hide bounds and window state in the Oracle-owned automation profile before moving the window off-screen.
- Visible runs restore only when that saved hide state exists; an ordinary visible or maximized window is left unchanged.
- Normal bounds are restored exactly, while maximized/fullscreen states are restored without forcing `windowState: normal`.
- A screen-aware fallback still recovers legacy Oracle-offscreen windows that have no saved marker.

Verification on the updated head:

- focused/adjacent browser tests: 108 passed
- `pnpm run check`: passed
- `pnpm run build`: passed
- full suite: 1,764 passed, 43 skipped
- disposable Chrome for Testing probe: exact saved-state round trip, ordinary visible no-op, and legacy recovery all confirmed

@clawsweeper re-review

IndelibleVivi:
Addressed both P1 findings in d7faf105.

- Fresh hidden launches now write the exact persisted profile placement marker before the off-screen launch flag reaches Chrome.
- Visible restoration now requires that Oracle-owned marker; the markerless geometry recovery path was removed.
- Added launch-order coverage and an unmarked negative-coordinate no-op regression test.
- Updated the PR body with a redacted built-code macOS lifecycle trace: exact pre-launch marker, clamped hidden bounds, exact restore, marker removal, and unchanged unmarked negative bounds.

Verification: 108 focused/adjacent browser tests passed; check and build passed; full suite passed with 1,764 tests and 43 skipped.

@clawsweeper re-review

PR #383, opened 2026-08-16
fix(browser): update Gemini web protocol and fallback handling

mpeter:
Update Gemini web request framing and current model headers, return clear errors for upstream failures/empty responses, and add `--no-gemini-fallback` while retaining fallback by default and through restarted sessions.

Maintainer repair preserves the model/thinking selectors and client UUID in the header as well as the new payload fields. It also keeps raw `gg-dl` image recovery reachable through the success guard. Oversized Google response headers now produce actionable Node startup guidance; the process's configured fetch/proxy/TLS transport is preserved.

Validation on `a25365a160db8952dde6a94aed2c87d191528520`:

- Typecheck, lint, build, docs/help checks, and full suite: 2,000 passed, 43 skipped.
- Real built CLI: text reply, PNG attachment interpretation, image generation, and image editing all succeeded with `--no-gemini-fallback` and the requested `gemini-3.1-pro` route.
- The host needed the documented `NODE_OPTIONS=--max-http-header-size=65536` for Google's session headers. No private dispatcher or new runtime dependency was introduced.
- Generated and edited images decode as 1024×1024 JPEGs; bytes/hash changed and the requested red triangle appears in the edit. These are synthetic test assets, inspected and stripped of metadata before upload.
- Final branch autoreview through P2 is scoped-clean. CI: https://github.com/steipete/oracle/actions/runs/34111459874.

Generated source (left), edited result (right):

Synthetic Gemini generation and image-edit proof

The live runs establish the requested route and pipeline behavior; they do not independently attest Google's backend model identity. Changelog credit is collected in the final notes PR.


steipete:
Final maintainer proof on `a25365a160db8952dde6a94aed2c87d191528520` is complete. All four CI jobs passed: https://github.com/steipete/oracle/actions/runs/34111459874.

The built CLI completed real Gemini text, PNG attachment interpretation, image generation, and image editing with `--no-gemini-fallback`, requesting `gemini-3.1-pro`. The documented `NODE_OPTIONS=--max-http-header-size=65536` accommodated Google's response headers without replacing the process's configured fetch/proxy/TLS transport. Both output images decoded as 1024×1024 JPEGs, their hashes differed, and the edit visibly added the requested red triangle. The inspected synthetic comparison is in the PR body.

Typecheck, lint, build, docs/help checks, and the full suite passed (2,000 tests; 43 skipped). Final branch autoreview through P2 is scoped-clean. The requested route and pipeline are proved; Google's actual backend model identity is not independently attested. Contributor credit is retained, and the changelog entry is collected in #462.

PR #386, opened 2026-08-16
fix(browser): bind Pro responses to committed turns

IndelibleVivi:
## What

- record a per-turn Pro receipt with dispatch time, response elapsed time, estimated input tokens, uploaded bytes, prompt digest, and committed user-turn index
- split send-attempt and durable-commit callbacks so recovery metadata cannot treat a click as proof of submission
- require every dispatch, timing, workload, commit, digest, and turn-index field whenever any new-format Pro marker is present
- verify the exact committed user turn before reattach accepts an answer, and treat partial or mismatched identity evidence as terminal
- preserve marker-free historical sessions and legitimate fast responses; elapsed time is durable evidence, not a default rejection threshold
- enforce receipt completeness at final session admission whenever either Pro config or a persisted Pro marker activates the contract

## Why

Pro effort selection already fails closed, but output acceptance did not have a durable trust boundary between sending a prompt and harvesting a response. Run-level elapsed time could include browser setup, follow-up turns could inherit the initial workload, and reattach could scope recovery by a prompt preview rather than a stored committed-turn identity.

This change binds timing, workload, and recovery to the same browser turn. Missing attachment sizes are resolved before dispatch, incomplete active-turn metadata is never replaced with an earlier turn workload, and a partial marker cannot bypass validation when configuration is stale or mis-resolved.

## Compatibility

Marker-free sessions created by older versions remain recoverable. A structurally complete fast Pro response remains valid regardless of elapsed time. Once any new-format marker exists, the full receipt is mandatory and incomplete evidence fails closed.

## Checks

- `pnpm run check`
- `pnpm run docs:check`
- `pnpm run build`
- `pnpm exec vitest run tests/browser` (49 files; 879 passed, 1 skipped)
- `pnpm test` (152 files passed, 18 skipped; 1787 tests passed, 43 skipped)

## Live browser evidence

Ran one useful direct-CDP GPT-5.6 Pro review against the six public source/test files after the two P1 fixes; this was not a dry run or a smoke-only prompt.

- model picker evidence: requested generic Pro target resolved to GPT-5.6 Sol, selected state verified
- initial capture timed out while the single assistant turn was still streaming and was stored as incomplete; no prompt was re-submitted
- `session --render` reconnected to the original conversation, rejected mid-stream capture, later confirmed terminal output, re-verified the committed prompt digest/index, saved one receipt, and marked the same session completed
- redacted completed receipt: commit=true; prompt digest present; logical/committed indices=0/0; elapsed=782645ms; input=249 estimated tokens plus 109860 attachment bytes; receipt count=1
- the live review confirmed both P1 fixes and identified a config-gated final-admission P2; commit `ad237ff9` adds the marker-gated admission check and its red-to-green regression test

No local paths, cookies, account data, conversation URL, or private files are included in this evidence.


IndelibleVivi:
Addressed both P1 findings in `ad237ff9`: any new-format marker now requires a complete committed-turn receipt, while structurally complete fast Pro responses remain compatible. A useful live GPT-5.6 Pro run exercised timeout → same-session reattach without resubmission; the completed receipt and redacted proof are in the PR body. The live review also found a config-gated final-admission P2, now covered by a red-to-green regression and fixed by marker-gated validation. Final checks: format/lint/typecheck/docs/build, 879 browser tests, and 1787 full-suite tests passed. @clawsweeper re-review

steipete:
Closing: this rewrites the receipt schema strictly and reworks recovery broadly, which carries real compatibility risk for existing receipts, without a bounded demonstration of the root cause it's meant to fix. If there's a specific failure you hit, a minimal reproduction plus a narrow fix would be much easier to land — and safer than a schema-wide change.

PR #389, opened 2026-08-16
fix(serve): stage remote attachments under unique paths

### LINKED ISSUE #387: Remote-host attachments sharing a sanitized name silently overwrite each other (post-sanitization basename collision)
**Version:** oracle 0.18.0 (`main` @ `083bba7e`), remote-host path (`ORACLE_REMOTE_HOST` / `browser.remoteHost`)

Related to #385, but a distinct mechanism on a different path. #385 is the local browser upload path, where `attachmentDataTransfer.ts` builds each browser `File` from `path.basename()`. This one happens server-side, before the browser is involved at all, and survives any client-side guard on original basenames.

## What happens

`sanitizeName` in `src/remote/server.ts:722` maps every character outside `[a-zA-Z0-9._-]` to `_`:

```ts
function sanitizeName(raw: string): string {
  return raw.replace(/[^a-zA-Z0-9._-]/g, "_");
}
```

At `src/remote/server.ts:229-238` that sanitized name becomes the write path, and each attachment entry then records that path:

```ts
const safeName = sanitizeName(attachment.fileName ?? `attachment-${index + 1}`);
const filePath = path.join(attachmentDir, safeName);
await writeFile(filePath, Buffer.from(attachment.contentBase64, "base64"));
attachments.push({ path: filePath, ... });
```

Two attachments whose names differ only in characters the regex folds therefore resolve to the same `filePath`. `writeFile` overwrites the first payload, and both pushed entries point at the surviving file. The fallback loop at `:249-258` has the same shape.

Concretely, `a b.txt` and `a_b.txt` both sanitize to `a_b.txt`.

## Why it is easy to miss

The names are genuinely distinct on the client, so any collision check on original basenames passes. The divergence only appears after sanitization, on the server, by which point the write has already happened. There is no warning and no count mismatch, so the run reports the full number of files.

## Impact

Same class as #385: silent content loss, with the surviving file wearing a plausible name. A caller sees the expected file count and a successful run while one attachment's content was never sent. The recurring-basename case that makes #385 painful in multi-directory packs applies here too, widened to any pair of names differing only in sanitized characters (spaces being the common one).

## Suggested fix

Collision-free path allocation at the write site, rather than a client-side check. Two reasons to prefer the server side:

1. A client-side guard on sanitized names requires the client to model the server's sanitization rule, which breaks the moment that regex changes.
2. Version-mismatched clients would never receive a client-side fix, whereas server-side allocation protects every caller.

Suffixing on collision (`a_b.txt`, `a_b-2.txt`) or allocating per-attachment subdirectories both work. Erroring is also defensible for consistency with the #385 guard, though the names here are legitimately distinct from the user's point of view, which argues for allocation over rejection.

## Not attempted

I did not run a live remote-host session against this; the report is from reading `main` at `083bba7e`. The mechanism is visible in source and the two loops are unambiguous, but a live confirmation would be worth having before anyone builds on it.

postoso:
Fixes #387.

Remote-host attachments whose names differ only in characters the server sanitizes are staged to the same path, so one payload overwrites the other before upload. `sanitizeName` maps everything outside `[a-zA-Z0-9._-]` to `_`, so `a b.txt` and `a_b.txt` both become `a_b.txt`, `writeFile` overwrites the first, and both attachment entries end up referencing the survivor. The run reports the full file count and succeeds.

Both staging loops had the defect: the primary loop and the fallback-submission loop.

## Approach

Each attachment stages under its own index-scoped directory:

```
attachments/<index>/<sanitized-basename>
fallback-attachments/<index>/<sanitized-basename>
```

Uniqueness is structural rather than a suffixing heuristic, and the sanitized basename is unchanged, so the filename the browser ultimately sees is exactly what it was before. Payload order is preserved.

## Real behavior proof

A real `oracle serve` instance on localhost, a real client run over HTTP with `--remote-host`, two files whose names collide only after sanitization, each carrying a distinct sentinel. The staging directory was captured during the run.

**Before (unmodified `main`):**

```
attachments/a_b.txt  ==>  SENTINEL-BETA-underscore-version
```

One file. The alpha payload is gone, overwritten by beta under the shared sanitized name. The client reported `files=2` and the run completed successfully in 34.0s, which is what makes this silent.

**After (this branch):**

```
attachments/0/a_b.txt  ==>  SENTINEL-ALPHA-space-version
attachments/1/a_b.txt  ==>  SENTINEL-BETA-underscore-version
```

Both payloads survive with distinct staged paths and the sanitized basename preserved. Client reported `files=2`, run completed in 39.9s.

Same harness, same two files, same transport; the only difference is the patch.

## Tests

`tests/remote/server.test.ts` gains coverage for colliding names in the primary loop, the same in the fallback loop, and a non-colliding control asserting ordinary behavior and order are unchanged. Each asserts distinct staged paths, expected sanitized basenames, preserved display paths, and that both distinct payloads survive on disk.

The fallback test fails without the fallback-loop half of the fix (`expected 1 to be 2`) and passes with it, so the loop that is easy to miss is covered by proof rather than inspection.

## Checks

`pnpm vitest run tests/remote/server.test.ts`, `pnpm lint`, `pnpm test` (1764 passed, 43 skipped), `pnpm build`, and `pnpm docs:check` all pass.

`pnpm check` reports a pre-existing `CHANGELOG.md` formatting issue that reproduces on an untouched `main` worktree at `083bba7e` and is not in `ci.yml`.

## Changed since first review

Dropped the `CHANGELOG.md` entry per review; changelog left to the release owner. The user-visible change is described above.

## Interaction worth flagging

While capturing the before case I ran it against a build carrying the #388 browser-side guard. On unmodified staging both attachments resolve to the same path, so `runBrowserMode` receives two entries with an identical basename and #388's boundary check rejects the run with a clear error rather than uploading a duplicated payload. The two fixes compose: #388 turns this failure loud at the browser boundary, and #389 removes the collision at its source so the run proceeds with both payloads intact.

### DIFF
--- a/src/remote/server.ts
+++ b/src/remote/server.ts
@@ -229,7 +229,9 @@ export async function createRemoteServer(
       const attachmentsPayload = Array.isArray(payload.attachments) ? payload.attachments : [];
       for (const [index, attachment] of attachmentsPayload.entries()) {
         const safeName = sanitizeName(attachment.fileName ?? `attachment-${index + 1}`);
-        const filePath = path.join(attachmentDir, safeName);
+        const stagingDir = path.join(attachmentDir, String(index));
+        await mkdir(stagingDir, { recursive: true });
+        const filePath = path.join(stagingDir, safeName);
         await writeFile(filePath, Buffer.from(attachment.contentBase64, "base64"));
         attachments.push({
           path: filePath,
@@ -247,7 +249,9 @@ export async function createRemoteServer(
           : [];
         for (const [index, attachment] of fallbackPayload.entries()) {
           const safeName = sanitizeName(attachment.fileName ?? `fallback-attachment-${index + 1}`);
-          const filePath = path.join(fallbackAttachmentDir, safeName);
+          const stagingDir = path.join(fallbackAttachmentDir, String(index));
+          await mkdir(stagingDir, { recursive: true });
+          const filePath = path.join(stagingDir, safeName);
           await writeFile(filePath, Buffer.from(attachment.contentBase64, "base64"));
           fallbackAttachments.push({
             path: filePath,
--- a/tests/remote/server.test.ts
+++ b/tests/remote/server.test.ts
@@ -149,6 +149,45 @@ describe("remote browser service", () => {
   );
 
   test.skipIf(!CAN_LISTEN_LOCALHOST)(
+    "stages colliding primary attachment names without losing payloads",
+    async () => {
+      await expectRemoteAttachmentStaging({
+        location: "primary",
+        files: [
+          { fileName: "a b.txt", content: "primary with space", stagedName: "a_b.txt" },
+          { fileName: "a_b.txt", content: "primary with underscore", stagedName: "a_b.txt" },
+        ],
+      });
+    },
+  );
+
+  test.skipIf(!CAN_LISTEN_LOCALHOST)(
+    "stages colliding fallback attachment names without losing payloads",
+    async () => {
+      await expectRemoteAttachmentStaging({
+        location: "fallback",
+        files: [
+          { fileName: "a b.txt", content: "fallback with space", stagedName: "a_b.txt" },
+          { fileName: "a_b.txt", content: "fallback with underscore", stagedName: "a_b.txt" },
+        ],
+      });
+    },
+  );
+
+  test.skipIf(!CAN_LISTEN_LOCALHOST)(
+    "preserves ordinary non-colliding attachment names and payload order",
+    async () => {
+      await expectRemoteAttachmentStaging({
+        location: "primary",
+        files: [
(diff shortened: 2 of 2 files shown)


steipete:
Superseded by #433 (merged) — a reviewed integration allocating distinct upload basenames in BOTH staging loops while preserving original display paths (the separate-directories approach here kept colliding basenames, which the browser runner's attachment-basename-collision guard rejects before Chrome starts). Verified with real Chrome preserving all payload bytes across eight files. Thank you @postoso — credit preserved in the landed commit.

postoso:
Thanks for landing it and for keeping the credit. Allocating distinct basenames in both staging loops is the better shape; the per-index directories I used would have kept tripping the collision guard on the browser side.

PR #396, opened 2026-08-18
fix(remote): say what a lost bridge connection means for the run

frontierkodiak:
## What

Killing `oracle serve` while a run was in flight left the caller with:

```
ERROR: aborted
```

That is the socket's description of the symptom, and it omits the part that changes what the operator should do: the run had **already been accepted**, so the browser work is not undone by losing the stream and the ChatGPT conversation very likely exists on the bridge host. Resubmitting — which is what "aborted" invites — opens a second conversation.

## Approach

The client now distinguishes the two transport failures it can have, using whether the service returned 200 (i.e. took the request) as the dividing line:

- **before acceptance** — the bridge is unreachable; say so and point at checking the service and port.
- **after acceptance** — name the bridge, say the run was in progress, and point at `oracle status` on that host to reattach rather than resubmit.

The underlying socket error is preserved as `cause` in both cases.

## Real behavior

Before, with the service killed 10s into a run:

```
ERROR: aborted
```

After, same scenario:

```
ERROR: Lost the connection to the research bridge at 127.0.0.1:9473 while the run
was in progress (aborted). The run may have started in ChatGPT before the
connection dropped: check `oracle status` on the bridge host for a session from
this run and reattach to it rather than resubmitting, which would create a second
conversation.
```

Exit code is 1 in both cases — this changes what the caller is told, not whether it fails.

## Tests

Two, driven against a stub that accepts and then drops the socket (the behaviour under test is entirely the client's, and a real server cannot drop mid-run without also waiting on the run it is pretending to perform): a lost-mid-run connection names the bridge and says reattach-not-resubmit, and an unreachable bridge is reported as unreachable rather than as a lost run.

Full suite green: 1763 passed / 43 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- Remote: say what a lost bridge connection means for the run. Losing the service mid-run surfaced the bare socket error ("aborted"), leaving out the consequential part — the run had already been accepted, so the ChatGPT conversation very likely exists on the bridge host and resubmitting opens a second one. The client now distinguishes a bridge it could not reach from one it lost mid-run, and the latter points at `oracle status` on the bridge host to reattach rather than resubmit.
+
 ## 0.18.0 — 2026-08-14
 
 ### Changed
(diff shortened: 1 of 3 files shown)


steipete:
Closing: the recovery guidance this adds points users at host-side `oracle status`, but remote runs don't create host-side sessions — so the suggested command won't show anything and sends people down a dead end. Guidance that reflects where the remote session actually lives (or that explicitly says a remote run has no host-side session to inspect) would be welcome.

PR #398, opened 2026-08-18
fix(serve)!: accept only conversation-scoped fields from remote clients

frontierkodiak:
> Security-relevant. Happy to split the banner fix out if you'd rather take them separately.

## What

`oracle serve` overrode six `browserConfig` fields it knew were dangerous and passed the rest of `BrowserSessionConfig` through to `runBrowserMode` verbatim. The remainder is not inert:

| field | what it does on the host |
| --- | --- |
| `chromePath` | names an executable this process spawns |
| `remoteChrome` | names a debugger to attach to |
| `copyProfileSource` | names a directory to copy a signed-in profile out of |
| `debugPort` | exposes one |
| `attachRunning` / `browserTabRef` | select an existing ChatGPT tab — an empty or `"current"` ref resolves to the first ChatGPT tab in the browser |

That makes a bridge token not a permission to ask ChatGPT a question but a permission to run code on the host. The single-flight guard hides the tab half today, since two runs never coexist; it does not hide the rest.

## Approach

Inverted to an allowlist. A client may describe the conversation it wants — URL, model, effort, archive mode, resume target, time budgets — and nothing about the machine. Rebuilding the config from that list rather than deleting known-bad keys also fixes the failure mode where a field added to `BrowserSessionConfig` later reaches Chrome the moment it exists.

The second commit is a smaller, related honesty fix: `oracle serve --host 127.0.0.1` advertised the host's LAN and tailnet addresses in its startup banner while bound to loopback only (verified with `lsof`, and by failing to connect from both). That banner is how an operator decides whether a port needs a tunnel or a firewall rule; for browser automation behind a bearer token, erring toward "more exposed than it is" is the wrong direction.

## Breaking

A remote caller that was setting host-scoped fields (`headless`, `hideWindow`, `chromePath`, `debugPort`, `attachRunning`, `browserTabRef`, cookie selection, or the shared-profile concurrency limits) is now ignored on those fields rather than obeyed. The host's own configuration decides them.

If you'd prefer these stay client-settable behind an explicit opt-in flag on the service (`--allow-client-host-config` or similar), say the word — the allowlist is a one-line escape hatch away from that shape.

## Tests

Three: the accepted set passes through, every host-scoped field is dropped (each named individually so a regression names its own hazard), and a missing config is treated as an empty one. Plus a banner test asserting a loopback bind advertises only loopback.

Full suite green: 1765 passed / 43 skipped.


frontierkodiak:
## Real behavior proof

A live `oracle serve` on loopback, and a raw client request carrying every host-scoped field the allowlist now drops:

```json
{
  "prompt": "Reply with exactly: allowlisted",
  "browserConfig": {
    "desiredModel": "gpt-5.5",
    "chatgptUrl": "https://chatgpt.com/g/g-p-.../project",
    "archiveConversations": "never",
    "chromePath": "/tmp/definitely-not-chrome",
    "browserTabRef": "current",
    "attachRunning": true,
    "headless": true,
    "copyProfileSource": "~/Library/Application Support/Google/Chrome",
    "debugPort": 9222
  }
}
```

Result:

```
RESULT answerText: 'allowlisted'
```

The run completed normally against the host's own Chrome. That is the proof: had `chromePath` been honoured the service would have tried to spawn `/tmp/definitely-not-chrome` and failed; had `copyProfileSource` been honoured it would have copied the operator's signed-in personal profile; had `browserTabRef: "current"` been honoured it would have driven whatever ChatGPT tab happened to be open. None of that happened, and the conversation-scoped fields in the same payload — model, project URL, archive mode — all took effect.

Before this change the same payload would have been obeyed field for field.

(Redacted: project id, bearer token, and profile path are host-specific.)

Full suite: 1765 passed / 43 skipped.

@clawsweeper re-review

PR #399, opened 2026-08-18
feat(browser): capture provider-native conversation evidence

frontierkodiak:
ChatGPT's copy-button Markdown can change LaTeX escapes and subscripts while still looking like a successful answer. This adds opt-in provider-native evidence so users can retain the exact conversation record and check the rendered answer against independently fetched text digests.

`--browser-capture-provider-native` saves the verbatim conversation JSON and an evidence file alongside session artifacts. A second request computes SHA-256 digests inside the authenticated page. Fidelity is bound to the captured assistant message on the active branch; prompt echoes, stale answers, and alternate branches cannot satisfy that comparison. Unsupported content and missing message IDs report `unknown`.

The capture uses Oracle's existing browser connection, has a 30-second total fetch/drain budget and an 8 MiB per-document ceiling, and preserves the normal copy-button/DOM answer on any failure. Tokens remain in the page; diagnostics use fixed reasons. Artifact writes are private on POSIX, and successful writes, typed failures, and fidelity survive in session metadata. The full record can contain prior turns, alternate branches, and provider metadata; documentation covers retention and export. User-config defaults and explicit opt-out work; project configs and remote bridge clients cannot enable full-conversation export.

During landing, the maintainer rebased onto current main, repaired failure/result/config plumbing, and replaced the unpublished Python-compatible fallback normalizer with the documented `text-fields-v1` format. Unknown/mixed content keeps its original bytes in the raw artifact without guessed text hashes. Deep Research can save the record while reporting unknown fidelity when no assistant message ID is available.

Independent autoreview is scoped-clean at P0–P2. Validation: full local suite (2,507 passed, 44 skipped), formatting/typecheck/lint, docs checks/site build, build, packed CLI smoke, and Oracle's connection-reuse proof. `node scripts/provider-native-capture-proof.mjs` exercises the compiled code with the contributor's recorded document and synthetic math responses, verifies raw bytes and independent hashes on disk, detects changed notation, checks challenge fallback, and verifies CLI flags. The compiled proof also runs in CI on Linux, macOS, and Windows. A single live Chrome attach was attempted through Oracle's reuse API but failed before sign-in could be checked; no live ChatGPT result is claimed.

Thanks @frontierkodiak (Caleb Sowers) for the proposal, implementation, and recorded fixture.


steipete:
Closing this feature proposal pending a maintainer decision on provider-native conversation export. Persisting full authenticated conversation documents needs a defined storage/retention contract and an explicit host-owned authorization boundary for any remote export. The current remote allowlist does not enable this flag, and adding it requires that decision first.

If adopted, I recommend an opt-in local-only first version, with authenticated artifact proof and corrected materialization metadata before considering remote access. I have retained that decision in the maintenance report; no contributor revision is requested.

steipete:
Reopening: this proposal is on an explicit maintainer hold pending Peter's decision on provider-native conversation export, and that hold is authoritative. The closure above was a triage error, not a decision. Nothing is requested from the contributor.

steipete:
Landed after rebasing onto current main and tightening message-bound fidelity, bounded failure handling, private artifact writes, config defaults, and persisted evidence status. I simplified the unpublished normalization format so unsupported content reports unknown instead of guessing text. The full local gate and exact-head CI passed, including compiled recorded/synthetic-fixture proof; the single live Chrome attach was unavailable. Thanks @frontierkodiak for the implementation and fixture!

PR #400, opened 2026-08-18
feat(serve): add opt-in bounded run admission and cancellation

frontierkodiak:
Integrate the opt-in bounded run queue with current main after attachment-context guards and thinking-effort evidence landed. The branches remain independent; contributor history and human credit are preserved.

The integration fixes cancellation cleanup: normal CDP calls still stop on abort, while an explicit cleanup scope can remove the caller's installed guards. Guard registration awaits are inside try/finally, so cancellation before the registration acknowledgement still removes listeners. Borrowed tabs survive; cancelled owned targets are closed.

Opt-in capacity is reserved when authenticated headers arrive, before buffering request bodies. Slow uploads retain FIFO admission order; invalid or disconnected bodies release reservations. Active/queued health counts include body reception. The individual payload-size policy is unchanged.

Plain serve retains its single-flight HTTP 409 behavior. Concurrent admission requires --max-concurrent-runs; host tab limits clamp capacity, --max-queued-runs bounds FIFO waiting, and full queues return 503. Existing legacy disconnect behavior and the request configuration allowlist remain intact.

Validation: 2,201 tests passed, 45 skipped; lint/build/docs passed. The full built service/client queue matrix, 27 attachment cases, and effort-evidence bridge matrix passed. The service matrix also verifies incomplete-body saturation, refusal before payload upload, and release of cancelled readers. The independently built #419 client completed against this service. Real CDP acknowledgement delays exercised local/remote file, prompt, plus and send cancellation; all listeners were removed, no prompt was submitted, and borrowed targets survived. A real built CLI/service case cancelled with a prompt guard installed and retired its owned target. Final branch review and exact-head CI are linked in the proof comment.

Limits: no signed-in ChatGPT concurrency/Pro/throttling proof or new native Windows/WSL guard-cancellation proof. Cancellation cannot revoke files already handed to the provider or guarantee that backend generation/billing stops. The remote file-transfer guard itself is synchronous; that case cancels its pending acknowledgement after the atomic renderer operation completes.

Deferred changelog (in #464): Remote: add opt-in bounded concurrent run admission with FIFO queuing, host-cap enforcement, cancellation, and isolated artifacts while preserving default single-flight HTTP 409 behavior; thanks @frontierkodiak.


frontierkodiak:
## Review follow-up

All three P1 findings were correct and are fixed in `db4c…`.

**Preserve the configured shared-browser tab cap.** You're right that the dependency was backwards. The tab cap is the physical constraint on a shared profile and belongs to the host; pinning it to whatever the service admits silently discards an operator's lower choice — and "lower to avoid account throttling" is precisely the case, since ChatGPT rate-limits well below what the transport can drive. The service now reads the host's configured `maxConcurrentTabs` and admits at most that many, logging when it clamps:

```
[serve] Admitting 3 concurrent run(s): the shared-profile tab cap (3) is lower than the requested 4.
```

**Forward AbortSignal to remote requests.** Also right, and the worst of the three: a caller aborting a remote run cancelled nothing while believing it had. The executor now destroys its HTTP request on abort — which is how the service learns to cancel, via its own disconnect handling — and refuses to send a request whose signal was already aborted.

**Make pre-connection setup cancellable.** The race was installed after the tab-lease wait, Chrome startup, and the CDP connection: the slowest stretch of a cold run and the one most likely to be waiting on a peer. It is now built before setup begins.

On your parenthetical about leaking — losing the race does not cancel the acquisition, so a lease granted *after* the caller gave up is now handed back explicitly. Otherwise cancelling during a queue wait burns a slot on the shared profile for the six-hour stale window, which is worse than not honouring the cancellation at all.

Two tests added for the bridge case: a caller aborting a remote run is observed as an abort inside the run, and an already-aborted caller never sends the request.

Full suite: 1774 passed / 43 skipped.

@clawsweeper re-review

steipete:
Maintainer proof for head fb9526f2d9ddbfb3c4e2a1e5c0f4abd195c77d7a.

The contributor history is preserved; current main was integrated without rewriting it. Plain serve still returns HTTP 409 while busy. Concurrent admission requires --max-concurrent-runs, with bounded FIFO queuing, host-cap clamping, and isolated host artifact directories. Programmatic cancellation is capability-negotiated; ordinary legacy disconnect behavior remains unchanged.

Validation: pnpm test (2,145 passed, 45 skipped), pnpm lint, pnpm build, and pnpm docs:check passed. The submitted RunSlots implementation reproduced two queued waiters with a configured bound of one; the atomic-admission regression now passes.

Real execution: node scripts/serve-queue-proof.mjs ran the built service and CLI clients against isolated real Chrome with synthetic pages. It passed default 409, effective host cap two despite requested capacity four, FIFO D then E, overflow 503, queued cancellation without submission, active cancellation with owned-target retirement in 1,323 ms, peer survival, distinct host sessions for identical client names, and prior-main client completion. Real integration also covered foreign lease preservation, cancellation during composer profile-lock waits, borrowed-target survival, delayed target creation and Chrome acquisition, and temporary Chrome cancellation with PID/profile removal. The fixture is included in Linux CI.

Local and final branch autoreview through P2 are clean. CI: https://github.com/steipete/oracle/actions/runs/34134556069 — all four jobs succeeded on the exact head.

Limits: no fresh signed-in ChatGPT concurrent/Pro workload or provider-throttling proof, no new native Windows/WSL queue-cancellation live proof, and no assertion that provider generation or billing stops on browser cancellation. All provider pages and service tokens used here were synthetic. The existing Windows shared-Chrome proof and Windows unit/build checks passed in CI.

Deferred release-note line remains in the PR body for a later notes PR. Ready for orchestrator review; not merged.

steipete:
Final integration verified at 194278a0b370423f3dd0fb9b299a45cf48106755 against main 176cec67c4efc897e357876fc0f630b6b3c32864. Contributor history and human credit are preserved.

All four exact-head CI jobs passed: https://github.com/steipete/oracle/actions/runs/34144580783. Local and final branch autoreview through P2 are clean.

Two integration defects were reproduced and repaired: aborted CDP calls prevented attachment-guard cleanup, and admission was reserved only after buffering request bodies. Cleanup now has an explicit non-cancelled scope, with guard-registration awaits inside try/finally. Opt-in admission reserves capacity at authenticated headers, before body reads; slow bodies preserve FIFO order, and invalid/disconnected readers release their reservations. Default legacy 409 and individual payload-size policy are unchanged.

Validation: 2,201 tests passed, 45 skipped; lint/build/docs passed. The full built CLI/service matrix passed default 409, host-cap clamping, FIFO, early 503 refusal for incomplete-body saturation, reader cleanup, queued/active cancellation, peer preservation, session isolation, actual lease/profile-lock waits, and late Chrome/target cleanup. The independent #419 binary completed against this service. The 27-case attachment and effort-evidence/bridge matrices also passed.

Seven real CDP guard-phase cases passed: six observed persistent guards active before abort; the remote file-transfer case delayed the acknowledgement after its atomic renderer operation. All tracked listeners were removed, no prompt was sent, and borrowed tabs survived. A real built CLI/service case cancelled with a prompt guard installed and closed only its owned target.

Additional combined validation: Git's merge simulation of current #419 and #400 was conflict-free. An isolated archive of that combined source, with #464's notes, passed 2,202 tests (45 skipped), lint/build/docs, real artifact export, effort-evidence, and guard-cancellation proof. No branch or PR was merged.

Limits remain: no fresh signed-in ChatGPT/Pro/throttling proof, no new native Windows/WSL guard-cancellation execution, and no guarantee of revoking provider-held files or stopping backend generation/billing. All provider pages and service tokens were synthetic. Merge #464 after #419 and #400.

PR #402, opened 2026-08-20
feat: add `oracle conversation export` (read-only ChatGPT conversation archive, Obsidian format)

2nd-Bird:
## The problem this solves

People who keep Codex / Claude Code session logs in an Obsidian vault get a compounding knowledge graph out of it. But the years of thinking they did in ChatGPT *before* agents existed sit outside that graph, and merging it in is miserable: long threads have scrolled out of context, the built-in export is one giant JSON, and copy-pasting hundreds of turns by hand doesn't scale.

Oracle already has the one ingredient needed to fix this — a signed-in ChatGPT Chrome tab reachable over CDP. This PR adds `oracle conversation export`, which uses that tab **read-only** to turn an existing conversation into Markdown, one note per query/answer pair, straight into a Git-tracked folder that Obsidian reads as-is. It also works as a plain, faithful ChatGPT archive for people who don't use Obsidian, and it is equally usable from Codex, Claude Code or a shell since it is just a CLI subcommand.

## What it does

- **Never sends, clicks or navigates.** The only CDP calls are `Runtime.evaluate` (two same-origin GETs) and `close()`; a test asserts exactly that, and another greps the in-page expression for `click`/`submit`/navigation/non-GET.
- `--source api` (default): in the attached tab, `fetch('/api/auth/session')` → bearer token → `fetch('/backend-api/conversation/<id>')`. That is the canonical JSON the ChatGPT UI renders from, so branches, `create_time`, model slugs, canvas (`canmore`) documents and thoughts-only turns are all visible in one shot, and `complete` is always `true` (the `current_node` path is fully walked). Any signed-in ChatGPT tab is enough — if `ref` doesn't match a live tab it falls back to any ChatGPT tab.
- `--source dom`: the virtualized-scroll DOM crawl with `turnIndex` gap detection, kept as an explicit legacy fallback (it is structurally blind to thoughts-only turns and can report false "incomplete").
- `--format json` (v2 records: one per turn with `messageIds`, `segments`, `hiddenNodes`, `attachments`, `createTime`, `model`, sha256), `markdown`, `raw` (untouched backend body), and `obsidian` (`--out <dir>`: `ChatGPT-<id8>/NNN-YYYY-MM-DD-turn-TTT.md`, one byte-exact note per Q/A exchange with `query_sha256`/`answer_sha256` provenance, plus `INDEX.md` with wikilinks; `--timezone`, `--captured`, `--folder`, `--force`).
- `--omit-text` keeps IDs/hashes only.

The subcommand flag is `--source`, not `--engine`, to avoid colliding with the root `--engine <api|browser>` option. README gains a short section framing the workflow; `docs/cli-reference.md` documents every flag and the note shape; CHANGELOG updated.

## Verification

- `pnpm test`: 155 files passed, 1823 tests (62 new across `tests/browser/conversationExport.test.ts`, `conversationApiExport.test.ts`, `tests/cli/conversationCommand.test.ts`, `conversationObsidian.test.ts`); `pnpm typecheck`, `oxlint`, `oxfmt --check`, `pnpm docs:check`, `pnpm build`, `git diff --check` all green.
- Real-device smoke on Windows Chrome over CDP against 12 conversations (20 – 562 turns, up to 3 MB of backend JSON): every export `complete:true`; user and assistant text compared byte-for-byte against the raw backend JSON (e.g. 281/281 queries and 279/279 answers on the largest); no-tab fallback and `--format obsidian` exercised on the same conversations. The DOM source on the same tab returned partial results with an honest `complete:false`, which is why `api` is the default.

## Notes for review

- `src/browser/conversationExport.ts` dynamic-imports the api module to avoid a static import cycle.
- The fetch expression is guarded to `chatgpt.com` / `chat.openai.com` origins.
- Happy to split the obsidian format into a follow-up PR if you'd rather land the export core first.

https://claude.ai/code/session_01Q2TDk7gPAY1tutSVGkhN32


steipete:
Thanks for the extensive archive work and validation. We're keeping Oracle core focused on consultations and their session artifacts, so we're closing the general conversation-archive and Obsidian export proposal as out of scope.

PR #404, opened 2026-08-21
fix(browser): honor thinking time in Deep Research

### LINKED ISSUE #382: Browser: --browser-thinking-time is silently ignored in Deep Research mode
## Summary

Is it intentional that `--browser-thinking-time` is ignored when
`--browser-research deep` is enabled?

On current `main`, both the local and remote browser paths contain the same
guard:

```ts
const deepResearch = config.researchMode === "deep";
const thinkingTime = config.thinkingTime;

// Handle thinking time selection if specified. Deep Research owns its own effort flow.
if (thinkingTime && !deepResearch) {
  await ensureThinkingTime(...);
}
```

As a result, this command accepts both options:

```bash
oracle \
  --engine browser \
  --model gpt-5.6-sol \
  --browser-thinking-time pro \
  --browser-research deep \
  -p "Run a small research task."
```

but the browser execution path does not call `ensureThinkingTime()` before
activating and submitting Deep Research.

## Question

Does ChatGPT Deep Research have its own fixed or server-controlled effort, such
that `--browser-thinking-time` is intentionally not applicable?

If so, would it be safer for Oracle to reject this option combination or emit a
clear warning instead of accepting and persisting a `thinkingTime` value that
is not applied?

If Deep Research can use the selected ChatGPT effort, should Oracle select and
verify the requested effort before enabling Deep Research and submitting the
prompt?

## Expected behavior

One of these contracts would make the behavior unambiguous:

1. **Effort is supported for Deep Research:** Oracle selects and verifies the
   requested effort, and a strict request such as `pro` fails closed before send
   if it cannot be confirmed.
2. **Effort is not applicable to Deep Research:** Oracle rejects or clearly
   warns about `--browser-thinking-time` when `--browser-research deep` is
   selected, and does not present the configured effort as effective.

The current behavior appears to accept the option while silently skipping it.

## Why this matters

For automation that requires an exact ChatGPT configuration, selecting
`GPT-5.6 Sol` and activating Deep Research is not enough to claim that the run
used `Pro` effort. The current session configuration can contain both
`thinkingTime: "pro"` and `researchMode: "deep"` without browser-side evidence
that Pro was selected or verified.

This is adjacent to the model/effort issues discussed in #182, #303, #353, and
#373, but I could not find an issue covering the interaction between
`browser-thinking-time` and Deep Research.

## Source inspection

- Repository revision:
  `083bba7e61f487ad3d99b42039d9f603f61dc4ff`
- Package version: `0.18.0`
- Relevant paths:
  - `src/browser/index.ts`, local browser flow near the
    `thinkingTime && !deepResearch` guard
  - `src/browser/index.ts`, remote browser flow with the same guard
  - `src/browser/actions/thinkingTime.ts`
  - `src/browser/actions/deepResearch.ts`

I have not modified Oracle, and I am reporting this as a source-level behavior
question rather than claiming a live UI reproduction. I can run a focused live
test or help prepare a patch once the intended contract is clear.

joebasrawi:
Fixes #382

Oracle skips the requested thinking time when Deep Research is enabled, so the saved config can say one thing while ChatGPT uses another.

This makes both browser paths select and verify the requested effort before Deep Research starts. I added regression tests for that combination.

I ran a live browser check on ChatGPT Pro using GPT-5.6 Sol, Pro effort, and Deep Research. It completed successfully. These were the relevant log lines:

```text
Model picker: GPT-5.6 Sol
[browser] Thinking time: Pro
Deep Research mode activated
Clicked send button
Deep Research completed (21s elapsed)
```

The targeted browser tests pass, 62 total. The full local test suite, build, lint, formatting, and type checks also pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Browser: detect a disabled ChatGPT effort tier (e.g. an exhausted Pro allotment) before clicking it, and report the account's own reset notice instead of a misleading "selection unverified" failure. Thanks @enieuwy!
+
 ## 0.17.3 — 2026-08-13
 
 **Highlight:** browser-mode answers and recovery are reliable again — no more
--- a/src/browser/index.ts
+++ b/src/browser/index.ts
@@ -204,6 +204,21 @@ export function classifyPreservedBrowserErrorForTest(
 // defaults to Standard effort. ensureThinkingTime() already handles the
 // "already-selected" case as a no-op, so always attempting it is safe.
 
+type BrowserConfigWithThinkingTime = Pick<
+  ResolvedBrowserConfig,
+  "researchMode" | "thinkingTime"
+> & {
+  thinkingTime: NonNullable<ResolvedBrowserConfig["thinkingTime"]>;
+};
+
+function shouldApplyThinkingTimeSelection(
+  config: Pick<ResolvedBrowserConfig, "researchMode" | "thinkingTime">,
+): config is BrowserConfigWithThinkingTime {
+  // Deep Research uses the same effort picker, so research mode must not
+  // suppress an explicitly configured thinking-time selection.
+  return config.thinkingTime !== undefined;
+}
+
 type ChatGptUiWarningType = "rate_limit" | "temporary_unavailable" | "auth_or_challenge";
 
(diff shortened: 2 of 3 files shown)


joebasrawi:
@clawsweeper re-review

Issue #381, opened 2026-08-15
browser: prompt composition can hang past --browser-input-timeout


lifeofgurpreet:

lifeofgurpreet:
Additional current-version evidence:

I also tested the documented direct-attach path to avoid launching another Chrome instance against the persistent profile:

```text
--browser-attach-running --remote-chrome 127.0.0.1:<CDP-port> \
--browser-tab <target-id>
```

The Chrome DevTools endpoint was reachable, and `/json/list` returned the exact requested ChatGPT page target. Oracle 0.18 nevertheless failed immediately with:

```text
No running browser with attach metadata matched 127.0.0.1:<CDP-port>.
Enable remote debugging in chrome://inspect/#remote-debugging first.
```

This makes the recovery path unavailable even though the browser is demonstrably reachable over CDP. It also suggests the error is checking Oracle-specific attach metadata rather than validating the supplied DevTools endpoint/target.

PR #409, opened 2026-08-23
fix(browser): preserve latest-turn identity in virtualized chats

pdurlej:
## Summary

- recognize committed prompts when ChatGPT virtualizes the visible turn count
- require a real latest-user-turn transition before accepting the stable-count path
- match recovery to the final non-empty follow-up in multi-turn browser sessions
- reconstruct the initial system-plus-user prefix used by the browser composer
- wait for an assistant answer paired with the expected latest user turn before harvest persists or prints output
- use the stored browser input timeout for bounded recovery hydration
- fail closed with a bounded error instead of returning an older answer

## Problem

In long reused conversations, ChatGPT can replace visible DOM turns while keeping the visible count stable. A submitted prompt may clear the composer and begin streaming while `turnsCount` remains equal to the baseline, producing a false `prompt-commit-timeout`.

The inverse race also exists during harvest: the new user turn may be visible while the assistant selector still points at the previous answer. A single snapshot can therefore return stale output.

## Fix

Before dispatch, capture the latest visible user turn. The stable-count commit path now requires:

- a known pre-submit user-turn state and a latest user turn matching the submitted prompt
- either changed normalized text or two known, distinct native `data-message-id` values
- a cleared composer, active Stop control, and conversation URL

An existing but empty user node is treated as unhydrated/unknown, not as a new conversation. A missing message ID, an ID that appears only after dispatch, or the same ID on both snapshots cannot prove a repeated prompt was committed and therefore fails closed. Raw message IDs stay in memory and are not persisted or logged.

Harvest now polls within the session's resolved `browser.config.inputTimeoutMs` and accepts output only when the assistant follows the latest user turn. This reuses the existing configured/default 60-second input budget instead of introducing a new hardcoded deadline. Single-turn recovery matches the full normalized initial prompt as a prefix so attachment context remains compatible. Multi-turn recovery selects the final non-empty configured follow-up and requires an exact normalized match; browser follow-ups are sent without attachments. If identity cannot be established, Oracle returns an explicit stale-output error and does not persist or print the candidate answer.

When stored run options contain a custom system prompt, initial-turn recovery reconstructs the same trimmed `system`, blank line, and trimmed user prompt prefix that browser prompt assembly sends. The current 0.18.0 CLI does not expose a `--system` flag; this preserves stored/programmatic compatibility rather than claiming a new CLI workflow.

## Verification

Exact head: `c3095c3f9f153b61e1009c9328c6f9b2bae8b24c`

- focused prompt/harvest suite: 29 passed
- full suite: 1771 passed, 48 skipped
- `pnpm run lint`
- `pnpm run build`
- `pnpm run docs:check`
- changed-file `oxfmt --check`
- `git diff --check`

Independent Claude Opus 5 review identified concrete false-positive states across the review cycles: repeated historical prompts, different prompts sharing the first 120 normalized characters, a short follow-up such as `Yes` matching an unrelated latest user turn, and a partial structural identity changing when an old node hydrates. Each accepted finding was reproduced as a red test and repaired. The structural repair now compares only two non-empty native message IDs through a pure tested helper; it does not use virtualized turn indexes. Two conditional concerns were resolved by source inspection: the browser runner uses the same trim-and-filter rule for follow-ups, and browser mode rejects an empty initial prompt before dispatch. One bounded structural repair run exhausted its tool-turn limit and is not counted as approval. A separate focused Opus 5 review of the final system-prefix repair completed with verified model and verdict `proceed`; its only residual request was real-DOM evidence for a CLI path that 0.18.0 does not currently expose.

An exact-head GPT-5.6 Sol repeated-turn proof completed without API fallback:

- one public `package.json` attachment was delivered inline
- the initial prompt and configured follow-up used the same exact text
- both turns returned exactly `UPSTREAM_REPEAT_TURN_OK` in 32.8 seconds
- model selection was verified as GPT-5.6 Sol, `promptSubmitted=true`, and both session/model statuses were `completed`
- no stale-output timeout occurred

On the preceding reviewed head, a distinct multi-turn run completed in 20.3 seconds and `session --harvest --write-output` then recovered the exact final follow-up after the original CDP port disappeared. The current focused suite preserves that recovery coverage.

The custom-system regression uses the exact browser composer ordering and separator, then adds trailing inline-file context to the harvested DOM text. It failed through the five-second stale-output boundary before the repair and now succeeds on the first harvest. No live `--system` receipt is claimed because that flag is not present in the current CLI.

The timeout regression keeps a stale DOM response for 6.25 seconds under a configured 10-second input timeout. The previous hardcoded deadline rejected it; the current head recovers the fresh turn on poll 26. A second regression configures 750 ms and proves a permanent mismatch still fails closed after four polls, so the repair remains bounded and operator-configurable.

The tab lease registry ended with zero leases and zero waiters.

## Compatibility

Historical sessions without a stored prompt still require a non-empty assistant answer paired with the latest visible user turn. When the stored prompt exists but the DOM cannot prove its full identity, harvest fails closed rather than returning potentially stale content.

PR #386 also touches prompt commit verification for Pro-specific durable receipts. This change is generic to ChatGPT DOM virtualization and CLI harvest freshness; it does not depend on the Pro receipt format, although a small conflict resolution may be needed if #386 lands first.

The exact upstream base currently reports unrelated `CHANGELOG.md` drift under whole-repository `format:check`; this PR does not touch that file, and all changed files pass the pinned formatter.

### DIFF
--- a/src/browser/actions/promptComposer.ts
+++ b/src/browser/actions/promptComposer.ts
@@ -903,6 +903,13 @@ async function verifyPromptCommitted(
     if (matchesPrompt && (baselineUnknown || info?.hasNewTurn)) {
       return typeof turnsCount === "number" && Number.isFinite(turnsCount) ? turnsCount : null;
     }
+    const virtualizedActiveCommit =
+      info?.composerCleared &&
+      info.inConversation &&
+      Boolean(info.lastMatched || (info.prefixMatched && info.stopVisible));
+    if (virtualizedActiveCommit) {
+      return typeof turnsCount === "number" && Number.isFinite(turnsCount) ? turnsCount : null;
+    }
     const fallbackCommit =
       info?.composerCleared &&
       Boolean(info?.hasNewTurn) &&
--- a/src/browser/actions/promptComposer.ts
+++ b/src/browser/actions/promptComposer.ts
@@ -257,12 +257,19 @@ async function readLatestUserTurnText(Runtime: ChromeClient["Runtime"]): Promise
         document.querySelectorAll('[data-message-author-role="user"], [data-turn="user"]'),
       );
       const latest = nodes[nodes.length - 1];
-      if (!latest) return '';
-      return latest.innerText ?? latest.textContent ?? '';
+      if (!latest) return { found: false, text: '' };
+      return { found: true, text: latest.innerText ?? latest.textContent ?? '' };
     })()`,
     returnByValue: true,
   });
-  return typeof result?.value === "string" ? result.value : null;
+  const value = result?.value as { found?: unknown; text?: unknown } | undefined;
+  if (value?.found === false) {
+    return "";
+  }
+  if (value?.found !== true || typeof value.text !== "string" || !value.text.trim()) {
+    return null;
+  }
+  return value.text;
 }
 
 export async function clearPromptComposer(Runtime: ChromeClient["Runtime"], logger: BrowserLogger) {
@@ -1031,6 +1038,7 @@ function summarizeCommitProbe(probe: CommitProbeState): Record<string, unknown> 
 // biome-ignore lint/style/useNamingConvention: test-only export used in vitest suite
 export const __test__ = {
   attemptSendButton,
+  readLatestUserTurnText,
   sendButtonTimeoutMs,
   verifyPromptCommitted,
 };
--- a/src/browser/actions/promptComposer.ts
+++ b/src/browser/actions/promptComposer.ts
@@ -213,6 +213,7 @@ export async function submitPrompt(
     );
   }
 
+  const baselineLastUserText = await readLatestUserTurnText(runtime).catch(() => null);
   const clicked = await attemptSendButton(
     runtime,
     input,
@@ -245,7 +246,23 @@ export async function submitPrompt(
     commitTimeoutMs,
     logger,
     deps.baselineTurns ?? undefined,
+    baselineLastUserText,
   );
+}
+
+async function readLatestUserTurnText(Runtime: ChromeClient["Runtime"]): Promise<string | null> {
+  const { result } = await Runtime.evaluate({
+    expression: `(() => {
+      const nodes = Array.from(
+        document.querySelectorAll('[data-message-author-role="user"], [data-turn="user"]'),
+      );
+      const latest = nodes[nodes.length - 1];
+      if (!latest) return '';
+      return latest.innerText ?? latest.textContent ?? '';
+    })()`,
+    returnByValue: true,
+  });
+  return typeof result?.value === "string" ? result.value : null;
 }
 
 export async function clearPromptComposer(Runtime: ChromeClient["Runtime"], logger: BrowserLogger) {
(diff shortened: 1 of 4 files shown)


pdurlej:
Addressed the P1 on exact head `750c829a9b1bb0799279a8bf8dc07e1a44d35a61`: recovery now selects the final non-empty configured browser follow-up, and follow-ups require an exact normalized latest-user match while the initial attachment-compatible path keeps prefix matching. Added red-to-green coverage for multi-turn recovery and short-prefix collisions. Focused 23/23, full 1765 passed with 48 skipped, lint/build/docs/format green. The PR body now includes an exact-head GPT-5.6 Sol multi-turn run plus successful recovery harvest after the original CDP port disappeared. @clawsweeper re-review

pdurlej:
Addressed the repeated-prompt P1 on exact head `5d597ef03de0445bd3e3a51b7ecee33b0cf5e652`. The stable-count path now accepts repeated text only when both pre/post snapshots expose non-empty, distinct native `data-message-id` values; virtualized turn indexes are not used, missing or newly hydrated IDs fail closed, and raw IDs are never persisted or logged. Added red-to-green coverage for changed/same/missing/late-hydrated IDs. Focused 27/27, full 1769 passed with 48 skipped, lint/build/docs/format green. An exact-head GPT-5.6 Sol browser run submitted an identical initial prompt and follow-up, captured both answers in 32.8s, completed both statuses, and left the lease registry empty. @clawsweeper re-review

pdurlej:
Addressed the custom-system compatibility P1 on exact head `d62b3ca7a5d011af92e0d6b038c41a25e227c4b9`. Initial-turn recovery now reconstructs the exact browser composer prefix from trimmed stored `system` plus `prompt`; follow-up matching is unchanged. The regression failed through the 5s freshness boundary before the patch and now succeeds on the first harvest, including trailing inline-file context. Focused 28/28, full 1770 passed with 48 skipped, lint/build/docs/format green; a bounded verified Claude Opus 5 review returned `proceed`. Accuracy note: Oracle 0.18.0 does not currently expose a `--system` CLI flag, so this is stored/programmatic compatibility and the PR does not claim a synthetic CLI live receipt. @clawsweeper re-review

steipete:
Closing: the hang you're addressing is real, but a hardcoded five-second harvest deadline is too aggressive — legitimate conversation hydration can exceed it on long threads or a slow account, so this trades a hang for spurious failures. Use the already-configured timeout rather than a new constant, and the fix becomes reviewable. (The prompt-preparation hang from #381 is separately fixed in #410, now merged.)

pdurlej:
@steipete Addressed on `c3095c3f9f153b61e1009c9328c6f9b2bae8b24c`: the hardcoded 5s deadline is gone. Harvest freshness now uses the session's existing resolved `browser.config.inputTimeoutMs` (60s default). A red-to-green regression keeps hydration stale for 6.25s under a configured 10s budget and then recovers; a second 750ms case still fails closed after four polls. Focused 29/29, full 1771 passed with 48 skipped, lint/build/docs/format green. The branch is updated without force-push. Would you reopen this PR for review?

PR #415, opened 2026-08-24
fix(browser): probe the DevTools endpoint when attach discovery finds nothing

### LINKED ISSUE #414: --browser-attach-running fails on Chrome 151: discovery needs a DevToolsActivePort file Chrome no longer writes
## Summary

`--browser-attach-running` cannot find a browser on Chrome 151, even when the DevTools endpoint is live and answering. Attach discovery looks for a `DevToolsActivePort` file in the user data directory, and Chrome 151 does not write one.

Every attach run fails with:

```
ERROR: No running browser with attach metadata matched 127.0.0.1:9222. Enable remote debugging in chrome://inspect/#remote-debugging first.
```

The message points at remote debugging, which is already enabled, so it sends you looking in the wrong place.

## Environment

- oracle 0.18.0
- Chrome 151.0.7922.174
- macOS, Apple silicon

## Reproduction

Start Chrome with remote debugging and a user data directory under the macOS discovery root:

```sh
"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  --remote-debugging-port=9222 \
  --user-data-dir="$HOME/Library/Application Support/AttachProfile" \
  --no-first-run --no-default-browser-check https://chatgpt.com/
```

Confirm the endpoint is up:

```sh
curl -s http://127.0.0.1:9222/json/version
# {"Browser":"Chrome/151.0.7922.174","Protocol-Version":"1.3", ... }
```

Chrome's own stderr also reports it:

```
DevTools listening on ws://127.0.0.1:9222/devtools/browser/609117bf-...
```

Then run any attach command:

```sh
oracle --engine browser --browser-attach-running -p "Reply with OK." -f some.jpg
# ERROR: No running browser with attach metadata matched 127.0.0.1:9222.
```

The user data directory contains no `DevToolsActivePort` file:

```sh
find "$HOME/Library/Application Support/AttachProfile" -maxdepth 3 -name DevToolsActivePort
# (no output)
```

## Cause

`resolveAttachRunningConnection` (`src/browser/attachRunning.ts`) calls `discoverDevToolsActivePortCandidates` (`src/browser/detect.ts:223`), which walks the roots from `resolveDevToolsActivePortDiscoveryRoots` (`src/browser/detect.ts:143`, `~/Library/Application Support` on macOS) looking for `DevToolsActivePort` files.

Discovery is therefore file-based. It never contacts the endpoint the user asked for. When Chrome does not write that file, `candidates` is empty and `attachRunning.ts:36` throws, regardless of whether `127.0.0.1:9222` is serving.

## Workaround

Writing the file by hand makes attach work immediately:

```sh
WS=$(curl -s http://127.0.0.1:9222/json/version | jq -r .webSocketDebuggerUrl)
printf '9222\n%s\n' "${WS#ws://*[0-9]}" > "$HOME/Library/Application Support/AttachProfile/DevToolsActivePort"
```

After that the same command connects and completes normally, including attachment upload and submission.

## Suggested direction

Two options, both small:

1. When an explicit host and port are known, probe `http://host:port/json/version` first and use `webSocketDebuggerUrl` from the response. Fall back to file discovery only when that fails. This removes the dependency on a Chrome implementation detail for the case where the user already told you where the browser is.
2. Keep file discovery, but on failure say what was actually looked for, for example "no DevToolsActivePort file found under `<root>`; the endpoint at 127.0.0.1:9222 was not probed". The current message sends people to `chrome://inspect`, which does not help.

Option 1 also fixes the case where the profile lives outside the platform discovery root, which is easy to hit with a throwaway profile in `/tmp`.

I am happy to send a PR for either if you have a preference.

devYRPauli:
Fixes #414.

## Problem

`resolveAttachRunningConnection` finds the browser only by file discovery. `discoverDevToolsActivePortCandidates` walks the platform root from `resolveDevToolsActivePortDiscoveryRoots` (`~/Library/Application Support` on macOS) looking for `DevToolsActivePort` files.

Chrome 151 does not write that file. So `candidates` is empty and attach throws, even when the endpoint is live. The old message also sends the user to `chrome://inspect`, where remote debugging is already enabled.

## Change

File discovery runs first and is unchanged. When it finds a candidate the behaviour is identical to before, including the `profileRoot` it returns.

Only when discovery finds nothing, read `webSocketDebuggerUrl` from `http://host:port/json/version` and attach to that. This is the case that fails outright today, so nothing that works now takes a new path.

`AttachRunningConnectionInfo.profileRoot` becomes `string | null`. The probe cannot supply a profile root, and the only consumer already accepts null: `src/browser/index.ts` assigns it to `config.remoteChromeProfileRoot`, which is typed `string | null`, defaults to `null`, and is read as `?? undefined`.

Two review findings are also fixed in `edf26a08`:

- **IPv6 hosts are bracketed before probing.** `parseRemoteChromeTarget` accepts `[::1]:9222` and stores the host as `::1`, so the probe built `http://::1:9222/json/version`. `new URL()` rejects that with `ERR_INVALID_URL`, so the fallback failed on the IPv6 path it exists to serve. The probe URL and the error message now use `formatWebSocketHost`, the helper `detect.ts` already had for this.
- **The abort timer is cleared on a rejected probe.** `clearTimeout` sat after the `await`, so a rejected fetch skipped it and left a live timer per attempt. It now runs in a `finally`.

When both discovery and the probe fail, the error names what was searched:

```
No running browser matched 127.0.0.1:63332. DevToolsActivePort discovery searched
/profiles/search-root, and endpoint probe http://127.0.0.1:63332/json/version failed:
connection refused.
```

## Real behaviour, Chrome 151.0.7922.174

Chrome started with a dedicated profile under `~/Library/Application Support`:

```
$ "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
    --remote-debugging-port=9222 --user-data-dir="<profile>" --no-first-run

$ curl -s http://127.0.0.1:9222/json/version
  Browser: Chrome/151.0.7922.174
  ws:      ws://127.0.0.1:9222/devtools/browser/<id>

$ find ~/Library/Application\ Support -maxdepth 3 -name DevToolsActivePort
  (no output)
```

The endpoint is live and no `DevToolsActivePort` file exists anywhere. That is the reported condition.

**Before, on `main`:**

```
$ oracle --engine browser --browser-attach-running --remote-chrome 127.0.0.1:9222 -p "..."
ERROR: No running browser with attach metadata matched 127.0.0.1:9222. Enable remote
debugging in chrome://inspect/#remote-debugging first.
```

**After, same browser, same command:**

```
$ oracle --engine browser --browser-attach-running --remote-chrome 127.0.0.1:9222 \
    -p "Reply with exactly: ATTACH OK"
[browser] Browser control: attach to an already-running local Chrome session
[browser] Browser guidance: Oracle opens a dedicated tab and leaves the existing browser process alone.
[browser] Thinking time: Pro
[browser] Archived ChatGPT conversation after saving local artifacts.
[browser] Model selection evidence: target=GPT-5.5; resolvedLabel=GPT-5.5; status=switched;
          strategy=select; verified=yes; source=chatgpt-model-picker
Answer:
ATTACH OK

22.8s - GPT-5.5[browser] - up 14 down 3
exit 0
```

Attach recovers, the model picker resolves, and the run completes.

## Tests

`vitest run tests/browser/attachRunning.test.ts`: 7 passed. Browser suite: 49 files, 882 passed, 1 skipped. `oxlint` clean, `tsc -p tsconfig.build.json --noEmit` exit 0.

Each new test fails against the code it guards:

```
probe removed        -> FAIL probes the endpoint when DevToolsActivePort discovery finds nothing
brackets removed     -> FAIL brackets an IPv6 host when probing the endpoint  (IPv4 control still passes)
finally removed      -> FAIL clears the abort timer after every rejected endpoint probe
```

Two pre-existing mocks in that file replaced the whole `detect` module, so they had to grow the new export. They now spread `importOriginal`, matching the other mocks in the file.

## One thing left alone

`src/browser/profileState.ts:317` (`verifyDevToolsReachable`) has both of the same defects: it builds `http://${host}:${port}/json/version` without brackets, and clears its timer after the `await`. That is pre-existing and outside this fix, so it is untouched here. Happy to send it as a follow-up.

### DIFF
--- a/src/browser/attachRunning.ts
+++ b/src/browser/attachRunning.ts
@@ -1,6 +1,7 @@
 import type { BrowserLogger, ResolvedBrowserConfig } from "./types.js";
 import {
   discoverDevToolsActivePortCandidates,
+  resolveDevToolsActivePortDiscoveryRoots,
   type DevToolsActivePortCandidate,
 } from "./detect.js";
 
@@ -8,7 +9,7 @@ export interface AttachRunningConnectionInfo {
   host: string;
   port: number;
   browserWSEndpoint: string;
-  profileRoot: string;
+  profileRoot: string | null;
 }
 
 export async function resolveAttachRunningConnection(
@@ -32,8 +33,18 @@ export async function resolveAttachRunningConnection(
     .sort(compareDevToolsCandidates);
 
   if (candidates.length === 0) {
+    const probe = await probeDevToolsBrowserWSEndpoint({ host, port });
+    if (probe.ok) {
+      return {
+        host,
+        port,
+        browserWSEndpoint: probe.browserWSEndpoint,
+        profileRoot: null,
+      };
+    }
+    const discoveryRoots = resolveDevToolsActivePortDiscoveryRoots();
     throw new Error(
-      `No running browser with attach metadata matched ${host}:${port}. Enable remote debugging in chrome://inspect/#remote-debugging first.`,
+      `No running browser matched ${host}:${port}. DevToolsActivePort discovery searched ${discoveryRoots.join(", ") || "no roots"}, and endpoint probe http://${host}:${port}/json/version failed: ${probe.error}.`,
     );
   }
   const candidate = candidates[0];
@@ -44,6 +55,47 @@ export async function resolveAttachRunningConnection(
     browserWSEndpoint: candidate.browserWSEndpoint,
     profileRoot: candidate.profileRoot,
   };
+}
+
+async function probeDevToolsBrowserWSEndpoint({
+  host,
+  port,
+  attempts = 2,
+  timeoutMs = 1000,
+}: {
+  host: string;
+  port: number;
+  attempts?: number;
+  timeoutMs?: number;
+}): Promise<
+  | { ok: true; browserWSEndpoint: string }
+  | { ok: false; error: string }
+> {
+  const versionUrl = `http://${host}:${port}/json/version`;
+  for (let attempt = 0; attempt < attempts; attempt++) {
+    try {
+      const controller = new AbortController();
+      const timeout = setTimeout(() => controller.abort(), timeoutMs);
+      const response = await fetch(versionUrl, { signal: controller.signal });
+      clearTimeout(timeout);
+      if (!response.ok) {
+        throw new Error(`HTTP ${response.status}`);
+      }
+      const version = (await response.json()) as { webSocketDebuggerUrl?: unknown };
+      if (typeof version.webSocketDebuggerUrl !== "string" || !version.webSocketDebuggerUrl) {
+        throw new Error("response did not include webSocketDebuggerUrl");
+      }
+      return { ok: true, browserWSEndpoint: version.webSocketDebuggerUrl };
+    } catch (error) {
+      if (attempt < attempts - 1) {
+        await new Promise((resolve) => setTimeout(resolve, 500 * (attempt + 1)));
+        continue;
+      }
+      const message = error instanceof Error ? error.message : String(error);
+      return { ok: false, error: message };
+    }
+  }
+  return { ok: false, error: "unreachable" };
 }
 
 function compareDevToolsCandidates(
--- a/src/browser/attachRunning.ts
+++ b/src/browser/attachRunning.ts
@@ -1,6 +1,7 @@
 import type { BrowserLogger, ResolvedBrowserConfig } from "./types.js";
 import {
   discoverDevToolsActivePortCandidates,
+  formatWebSocketHost,
   resolveDevToolsActivePortDiscoveryRoots,
   type DevToolsActivePortCandidate,
 } from "./detect.js";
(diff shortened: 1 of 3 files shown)


devYRPauli:
Both code findings are fixed in edf26a08, and the real-behaviour proof is now in the PR body.

The transcript is a real Chrome 151.0.7922.174 session: remote debugging on, `/json/version` live, and no `DevToolsActivePort` file anywhere under `~/Library/Application Support`. On `main` attach fails with "No running browser with attach metadata matched". With the fix the same command against the same browser attaches, resolves the model picker, returns its answer and exits 0 in 22.8s.

On the IPv6 finding: `parseRemoteChromeTarget` strips the brackets and stores `::1`, so the probe built `http://::1:9222/json/version`, which `new URL()` rejects with `ERR_INVALID_URL`. The URL and the error message now use `formatWebSocketHost` from `detect.ts`, which already existed for this and only needed exporting.

On the timer: `clearTimeout` moved into a `finally`.

7 tests pass. Removing the brackets fails only the IPv6 test and leaves the IPv4 control green; removing the `finally` fails only the timer test.

One related note, not changed here: `src/browser/profileState.ts:317` (`verifyDevToolsReachable`) has both of the same defects, the unbracketed URL and the timer cleared after the await. It is pre-existing, so I left it out to keep this PR scoped. Say the word and I will send it as a follow-up.

devYRPauli:
@clawsweeper re-review

The real-behaviour proof is in the PR body now. The review that completed at 16:31 started at 16:08, before the body carried it, so it still reads as mock-only.

The transcript is a real Chrome 151.0.7922.174 session with remote debugging on, `/json/version` live, and no `DevToolsActivePort` file anywhere under `~/Library/Application Support`. `main` fails with "No running browser with attach metadata matched". The same command against the same browser, with this branch, attaches, resolves the model picker, returns its answer and exits 0 in 22.8s.

steipete:
Superseded by #426 (merged) — a reviewed integration of this fallback that also bounds the full request lifecycle including body reads inside the retry budget (the original head could stall past the 2.5s budget on a slow body). IPv6 handling and timer cleanup preserved. Thank you @devYRPauli — co-authorship preserved on the landed commit.

PR #407, opened 2026-08-22
fix(browser): support Japanese archive controls

moeuu:
ChatGPT's Japanese UI labels the conversation menu as `その他` and the archive action as `アーカイブ`. The existing matcher only covers English and Polish, so browser runs with a Japanese UI cannot reach the archive action.

This adds the current Japanese menu and action labels, keeps `アーカイブを解除する` out of archive matches, and recognizes the localized success message.

Live browser check (2026-08-22):
- session: `japanese-archive-live-proof-success`
- Japanese UI: `lang=ja-JP`, conversation menu `その他`
- archive metadata: `attempted=true`, `archived=true`
- reopening the conversation showed the archived banner and `アーカイブを解除する`

Redacted output:

```text
[browser] Saved transcript artifact
[browser] Archived ChatGPT conversation after saving local artifacts.
Answer:
CHECK_ARCHIVE_OK
```

`oracle session japanese-archive-live-proof-success --render` also returned `CHECK_ARCHIVE_OK` from the saved transcript. The current account offers GPT-5.6 Sol rather than GPT-5.5, so the smoke test kept the model already selected in ChatGPT.

Tests:
- `pnpm check`
- `pnpm test`
- `pnpm build`
- `pnpm docs:check`

PR #416, opened 2026-08-25
test(browser): guard archive expression locale labels against silent drift

OfficialAbhinavSingh:
## Summary

- Add a locale-label drift guard for the ChatGPT archive-conversation matcher (`buildArchiveConversationExpression`).
- Test-only. No production code changed.

## Problem

`docs/browser-mode.md` names this gap directly:

> Model picker drift – we rely on heuristics… Consider snapshot tests or a small "self check" command.

The archive matcher is the clearest recent example. It carries four separate per-locale label lists (menu "more options", archive item/confirm, unarchive/restore exclusion, confirmation toast), each independently extended locale-by-locale (#407 added Japanese; Polish predates that). Nothing asserted the full set together — only a handful of Japanese literals were spot-checked. A future PR touching any one of these `.includes(...)` chains could silently drop or mistype an existing locale's label, and it would only surface later as a live-account regression report.

## Change

`tests/browser/archiveConversation.test.ts`: one inventory test per matcher (menu, archive action, unarchive exclusion, confirmation toast), asserting every locale label currently in `buildArchiveConversationExpression` is present. A drop or typo in any of them now fails CI at review time instead of live.

## Update: scoped assertions per matcher (addresses review feedback)

The first version checked each label against the *whole* serialized expression. Several labels ("archive", "アーカイブ") are legitimately repeated across `findArchiveMenuItem` and `findArchiveConfirmationButton`, so a global `toContain(label)` stayed green even when a label was deleted from just one of them — reviewed and flagged correctly.

Fixed: the expression is now sliced per matcher function before asserting, so a check can only be satisfied by that matcher's own labels. Labels are also matched as quoted literals (`'label'`) rather than bare substrings — "アーカイブ" is itself a substring of "アーカイブする", so a bare check would still pass after the standalone literal is removed as long as the longer one survives.

Verified against exactly the flagged scenario: removed the `'アーカイブ'` exact-match line from `findArchiveConfirmationButton` only (left it in `findArchiveMenuItem`). First attempt (unquoted slice check) stayed green — a false negative, confirming the substring issue above. Fixed to quoted matching, reran: failed on exactly that label. Restored the source, confirmed green again (14/14).

Original drift-catch proof (still holds against the current version): temporarily removed one Polish literal (`opcje`), reran, restored:

```
FAIL  tests/browser/archiveConversation.test.ts > archive expression locale label inventory > conversation menu ('more options') labels are present
AssertionError: ...to contain '\'opcje\''
```

`git diff --stat src/browser/actions/archiveConversation.ts` → no diff after both restores; production code was never touched.

## Checks

- `pnpm vitest run tests/browser/archiveConversation.test.ts` — 14 passed
- `pnpm test` — 152 files / 1795 passed / 44 skipped (unchanged skip count)
- `pnpm run typecheck` — clean
- `npx oxlint tests/browser/archiveConversation.test.ts` — clean
- `npx oxfmt --check` — clean

`CHANGELOG.md` left as-is (mis-formatted on main; release owner per #349).

## Scope note

This is a pilot on one file to establish the pattern before extending it — `thinkingTime.ts` (behind #405's effort-label locale fixes) and the other action files with per-locale matchers are natural next candidates, as follow-up PRs.

### DIFF
--- a/tests/browser/archiveConversation.test.ts
+++ b/tests/browser/archiveConversation.test.ts
@@ -171,12 +171,43 @@ describe("archiveChatGptConversation", () => {
   });
 });
 
-// Every locale label this matcher relies on, one line per literal. A future PR that
-// silently drops or renames one of these (while adding a new locale, for example)
-// fails here instead of only showing up as a live-account regression report — see
-// docs/browser-mode.md's "self check" ask and the localized-controls history in #407/#405.
+// Every locale label this matcher relies on, one line per literal, checked against the
+// SPECIFIC matcher function that consumes it (not the whole serialized expression) — a
+// global `expression.toContain(label)` would still pass if a label were deleted from one
+// matcher while an identical label happened to survive in a sibling matcher, since several
+// of these labels (e.g. "archive", "アーカイブ") are deliberately repeated across matchers.
+// Labels are matched as quoted literals (`'label'`), not bare substrings: "アーカイブ" is
+// itself a substring of "アーカイブする", so a bare check would stay green even after the
+// standalone "アーカイブ" literal is deleted, as long as "アーカイブする" still exists.
+// A future PR that silently drops or renames a label now fails here instead of only
+// showing up as a live-account regression report — see docs/browser-mode.md's "self check"
+// ask and the localized-controls history in #407/#405.
 describe("archive expression locale label inventory", () => {
   const expression = buildArchiveConversationExpressionForTest();
+  const quoted = (label: string) => `'${label}'`;
+
+  // Slice out one matcher function's own source, bounded by the next function's start,
+  // so an assertion against the slice can only be satisfied by that matcher's own labels.
+  const matcherSlice = (startMarker: string, endMarker: string): string => {
+    const start = expression.indexOf(startMarker);
+    if (start === -1) {
+      throw new Error(`matcher start marker not found in expression: ${startMarker}`);
+    }
+    const end = expression.indexOf(endMarker, start + startMarker.length);
+    if (end === -1) {
+      throw new Error(`matcher end marker not found in expression: ${endMarker}`);
+    }
+    return expression.slice(start, end);
+  };
+
+  const menuButtonSlice = matcherSlice("findConversationMenuButton", "visibleMenuCandidates");
+  const archiveItemSlice = matcherSlice("findArchiveMenuItem", "findArchiveConfirmationButton");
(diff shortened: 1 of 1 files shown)


steipete:
Maintainer triage: **LAND recommended** for this test-only patch, applied to current main `bbc1b3b0`. No production changes or additional patch were needed.

The scoped matcher slices and quoted literals address the earlier false-positive concern: a label in another matcher, or inside a longer label, cannot satisfy the intended inventory assertion.

Validation:
- `node_modules/.bin/tsc -p tsconfig.build.json` — passed.
- `node_modules/.bin/vitest run tests/browser/archiveConversation.test.ts` — 14 passed.
- `node dist/bin/oracle-cli.js --version` — `0.18.0`.
- `node dist/bin/oracle-cli.js --render -p 'Review the package metadata for release risks' --file package.json` — rendered the prompt and numbered file bundle successfully.
- The unchanged production archive expression ran in real Chrome against synthetic English, Japanese, and Polish controls: each returned `status: archived`, exactly one archive click.
- Separately, a synthetic signed-in ChatGPT Pro conversation returned `ORACLE_ARCHIVE_TRIAGE_PROOF`; the live header `More` → `Archive` flow returned to the home composer. The prior effort selection was restored. This was UI verification, not an Oracle CDP transport smoke; the localized fixtures were synthetic.
- Codex autoreview: no accepted/actionable P0 findings in the patch scope.

The available PR check is GitGuardian only; the orchestrator still owns normal CI and landing. Nothing was merged.

steipete:
Merged as 955363e028a4fdbcde6bdc2d318056d2105cd4b9. Thanks @OfficialAbhinavSingh for the scoped archive-label guards and the follow-up repair!

Landing verification on exact PR head `01b0a6a158f837804a76a127b2c6b44226303b41`:
- `node_modules/.bin/vitest run tests/browser/archiveConversation.test.ts` — 14 passed.
- Codex autoreview — no accepted/actionable findings at the configured P0 threshold.
- [CI run 32836279805](https://github.com/steipete/oracle/actions/runs/32836279805) — approved the pending fork run; Linux, macOS, Windows, and the real Chromium CDP proof all passed. GitGuardian also passed.

The earlier maintainer proof above covers the archive expression in real Chrome with synthetic English, Japanese, and Polish controls, plus the separately identified signed-in UI smoke. No additional live-browser run was needed for this unchanged test-only head. No changelog entry because production behavior is unchanged.

PR #420, opened 2026-08-26
fix(browser): bundle multiple text uploads

gwelinder:
## Summary

- Automatically combine multiple uploaded text files into one ordered, source-labelled text bundle while keeping a single text file under its original attachment name.
- Preserve explicit attachment/bundle modes and formats. Mixed raw inputs remain byte-preserving direct uploads unless explicit bundling or the platform attachment limit requires ZIP.
- Include expected names, missing UI/input names, send state, upload state, file count, and observed input names in direct-upload timeout errors.

## Root cause

The automatic policy bundled only after the 10-attachment platform limit. Nine Markdown files therefore took nine sequential upload/readiness paths. Completion correctly required all expected names and a ready send control, but if the ninth attachment never appeared, Oracle waited out the entire budget and then returned a generic timeout with no stalled filename or final readiness state.

The new boundary is semantic rather than incident-specific: one text file remains one attachment; multiple text files form one lossless labelled corpus.

## Validation

- `pnpm run check`
- `pnpm vitest run tests/browser/policies.test.ts tests/browser/prompt.test.ts tests/browser/attachmentsCompletion.test.ts tests/browser/pageActions.test.ts tests/browser/promptComposerExpressions.test.ts tests/browser/index.test.ts tests/browser/sessionRunner.test.ts` — 277 passed, 1 skipped
- `pnpm test` — 1,795 passed, 43 skipped
- `pnpm run build`
- `pnpm docs:check`

The mocked browser regression reproduces an 8-of-9 Markdown readiness stall, verifies the exact missing ninth filename/state, and proves the generated one-bundle upload reaches ready state. No Chrome smoke was run because Frog 20260826151633 explicitly requires this fix without launching Chrome.

Frog: 20260826151633


steipete:
Thanks for digging into the multi-file readiness stall and adding more useful missing-filename diagnostics.

I'm closing this version because it overlaps the upload-policy discussion in #406 and changes the default behavior for multiple text files. It also eagerly creates fallback bundles that remain unused after successful inline submissions, without a production cleanup owner. The mocked regressions are useful, but there is no real browser proof for the changed upload path.

Please bring the default-bundling proposal to the discussion in https://github.com/steipete/oracle/pull/406 so we can agree on one policy. A focused follow-up for readiness diagnostics or a current-main upload failure is welcome, with real browser evidence and lazy creation or verified cleanup for generated bundles. Thanks again for the contribution.

PR #432, opened 2026-08-28
fix(browser): recover no-op image prompt sends

oraclexing:
## Summary

- recover when ChatGPT leaves an attachment-bearing image prompt staged after a trusted send-button click
- compare canonical composer whitespace before using a single Enter fallback, and suppress the fallback when a new turn or stop control is already visible
- persist `promptSubmitted` only after the user turn is observable
- document the signed-in image-generation regression smoke for browser changes

## Root cause

On the current Windows ChatGPT home composer, an uploaded reference image could leave the visible enabled `data-testid="send-button"` as a no-op for CDP mouse input. Oracle already had enough evidence to see that the prompt stayed staged, but a raw-string comparison would miss it because ProseMirror renders paragraph boundaries with extra newlines. The run then waited the full prompt-commit timeout with zero conversation turns.

The fallback now canonicalizes whitespace, waits briefly for the click to commit, checks that no new turn or generation stop control appeared, and only then presses Enter once.

## Tests

- `pnpm vitest run tests/browser/promptComposer.test.ts tests/browser/promptComposerExpressions.test.ts` — 63 passed
- `pnpm run typecheck`
- `pnpm test` — 153 files passed, 18 skipped; 1871 tests passed, 53 skipped
- `pnpm run build`

## Signed-in browser smokes

Underlying browser-path reproduction and fallback evidence:

- Session: `chatgpt-image-fallback-smoke-two`
- GPT-5.6 Sol, Instant, one real PNG reference, attachment mode `always`
- Observed log: `Send click left the prompt staged; submitting once via Enter`
- Runtime: `promptSubmitted=true`, conversation URL changed from `/` to `/c/...`
- Live DOM: empty composer plus the exact submitted user turn
- Completed in 1m44s and downloaded two valid 1254x1254 PNG artifacts under `~/.oracle/generated/`

Exact MCP wrapper smoke through a fresh stdio `oracle-mcp.js` process:

- Tool: `chatgpt_image`
- Session: `chatgpt-image-mcp-exact-smoke`
- Runtime: `promptSubmitted=true`, conversation URL `/c/6a91962b-9810-83ee-9f62-8d517a5bd878`
- Status: `completed` in 1m20s
- `structuredContent.images` returned the requested output path
- Saved PNG independently validated as 1254x1254, 817,223 bytes, with a valid PNG signature
- Oracle browser slot released; retained Chrome/profile lock cleanup verified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Browser: recover when ChatGPT leaves an attachment-bearing image prompt staged after a no-op send-button click by submitting it once via Enter; normalize ProseMirror paragraph whitespace and persist `promptSubmitted` only after the user turn is observable.
 - Browser: attach to running Chrome when `DevToolsActivePort` metadata is absent, with IPv6 support and bounded endpoint retries that include response-body reads. Fixes #414. Thanks @devYRPauli!
 - Browser: recognize collision-renamed attachment chips, including short filenames, while keeping Unicode filename boundaries and visible extensions distinct across upload and send checks. Fixes #393. Thanks @devYRPauli!
 - Browser: select and verify thinking effort in ChatGPT's direct-slider picker without an Advanced submenu, keeping explicit Pro requests fail-closed. Fixes #422.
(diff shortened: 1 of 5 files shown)


steipete:
Thanks for the detailed reproduction and the signed-in browser/MCP smokes; that evidence is useful for understanding the no-op image send.

I'm closing this version because an unchanged composer after 750 ms cannot establish that the first click failed. A successful click can still be committing with no user turn or stop control visible, so the Enter fallback can submit the same image request twice. Main already improves activation and single-send targeting in #434 while keeping ambiguous commit timeouts out of retries.

A focused follow-up is welcome if the no-op still reproduces on current main. Please include delayed-success coverage and real browser evidence, and preserve one submission attempt unless there is reliable evidence that the first attempt failed. Thanks again for investigating this path.

PR #439, opened 2026-09-01
feat(browser): add Perplexity browser provider

mkubenka:
Drive perplexity.ai from browser mode alongside ChatGPT and Gemini web. `--model perplexity` runs Search and `--model perplexity-research` runs Deep research against the signed-in session in the persistent Oracle Chrome profile. The provider plugs into the existing executeBrowser seam, so the ChatGPT automation in browser/index.ts is untouched.

Answers carry their citations: sources are collected from both the inline links and the sources list, de-duplicated by URL, and appended to answerMarkdown. Files follow the usual prompt assembly, with text pasted inline and media uploaded to the composer. Image prompts save through --generate-image/--output, and --browser-follow-up continues the same conversation as a multi-turn transcript.

Bare `perplexity` ids are browser-only, so --engine api rejects them with a pointer to browser mode while OpenRouter ids such as `perplexity/sonar-pro` keep working unchanged. --remote-host is rejected up front because the remote browser service drives ChatGPT automation only and would otherwise answer a Perplexity request as ChatGPT. Dispatch is wired for both the CLI and the MCP consult tool.

Several page behaviours are load-bearing and documented in docs/perplexity.md so a later selector refresh does not undo them: completion must be scoped to the current turn and hold a steady length because the Stop button clears before the text settles, an image answer renders no prose turn at all, the sources list fills in after the answer, uploads only register through the intercepted file chooser, and the mode menu needs real CDP input events.

Lift the shared Chrome/CDP session handling out of gemini-web into browser/webSessionManager.ts so both web providers use one implementation rather than a third copy.

Proof:

- 105 focused tests pass across the Perplexity provider, executor, model resolution, follow-up resolution, and run-option routing (five files). Full suite: 1,959 passed / 43 skipped; 157 files passed / 18 skipped (26.35 s).
- Formatting, typecheck/lint, build, docs/help consistency (77 flags / six files), docs-site build, and the packed built-CLI help smoke all pass.
- Real Chrome CDP disconnect proof passes both cases: recoverable client disconnect with the target alive, and closed-window with no stale auto-reattach loop.

Live provider runs against a signed-in Perplexity account, through the built CLI and the MCP `consult` tool:

- Search (`--model perplexity`): answered in 11.9–15.1 s, returning 11–12 de-duplicated sources under a `## Sources` heading. The same run through MCP returned 11 sources in its result payload.
- Deep research (`--model perplexity-research`): the composer mode switches to Deep research and the answer completes in 53–67 s, returning 15–20 sources.
- Image generation: a descriptive prompt produced a 1536×1024 image; `--generate-image` wrote 2,409,028 bytes to disk, and the MCP path wrote its own image and persisted it as a session artifact (`image`, alongside `transcript`) so the MCP `images` response includes it.
- Attachments: a text file is pasted inline by the existing prompt assembly (input token count rose from ~20 to 645 for one source file). A `.png` uploads through the intercepted file chooser and the answer describes the image contents; two files upload in one chooser open and the answer names the subject of each.
- Follow-ups (`--browser-follow-up`): the second turn runs in the same conversation and resolves a pronoun against the first answer, and the result is returned as the standard multi-turn transcript.

Redacted trace of a signed-in Search run through the built CLI (`node dist/bin/oracle-cli.js --engine browser --model perplexity --verbose`). Account identifiers and the conversation id are omitted; nothing else is edited.

```
[perplexity-web] Starting Perplexity web executor
[perplexity-web] Reusing Chrome on port <port> for Perplexity (perplexity).
[perplexity-web] Navigating to https://www.perplexity.ai/...
[perplexity-web] Waiting for Perplexity UI to load...
[perplexity-web] Mode "Search" already active.
[perplexity-web] Typing prompt...
[perplexity-web] Sending prompt...
[perplexity-web] Waiting for the Perplexity answer...
[perplexity-web] Still generating... (waiting)
[perplexity-web] Completed in 15116ms (268 chars, 10 sources, 0 images).

Answer:
The current Node.js LTS release lines are v24 "Krypton" (Active LTS) and v22 "Jod"
(Maintenance LTS); v24 is the normal choice for new production deployments.
Node.js v26 is currently a Current release and is expected to enter LTS in October 2026.

## Sources

1. [endoflife](https://endoflife.date/nodejs)
2. [Node.js 24.20.0 (LTS)](https://nodejs.org/en/blog/release/v24.20.0)
3. [Download & Update Node.js to the Latest Version! Node v26 ...](https://blog.risingstack.com/update-node-js-latest-version/)
4. [nodejs/node: Node.js JavaScript runtime](https://github.com/NODEJS/NODE/)

15.1s · perplexity[browser] · ↑22 ↓67 ↻0 Δ89
```

The same run through the MCP `consult` tool returns `status: "completed"` with the answer and its `## Sources` block in `output`, a `perplexity` entry in `models`, and a `transcript` artifact; an image run additionally returns an `image` artifact for the file it wrote.

Routing guards, each exercised through the CLI:

- `--engine api --model perplexity` is rejected as browser-only; `--engine api --model perplexity/sonar-pro` still resolves to the OpenRouter model unchanged.
- `--engine browser --model perplexity/sonar-pro` is rejected: qualified ids stay on the API path instead of opening the web UI.
- `--engine api --model perplexity-labs` passes through to the custom-model path unchanged, and `--engine browser --model perplexity-labs` is rejected by name; only the documented aliases select the web provider.
- `--models gpt-5.5,perplexity` is rejected; multi-model runs force the API engine, so every requested id is checked.
- `--remote-host` with a Perplexity model is rejected, because the remote service drives ChatGPT automation only.
- `oracle --followup <perplexity-session>` no longer rebuilds a ChatGPT conversation URL from a Perplexity conversation id.

Scope and limitations:

- The live evidence above is manual, run from a personal signed-in Perplexity account; account identifiers and conversation ids are omitted. The committed tests mock the Chrome session and provider flow, so they guard the completion, upload, submit, and source-extraction logic but not the live selectors.
- There is no `tests/live/perplexity-live.test.ts` yet, so selector drift is not caught automatically. That is the main follow-up.
- `--browser-attachments always` (forcing upload for text files) is implemented but has not been exercised end to end.
- Session reattach and detached `--no-wait` runs have not been verified for this provider.
- Cloudflare can present a bot check on this domain. The provider detects it during both UI wait and response polling and fails fast with an actionable message rather than burning the response timeout, but that specific failure path is covered by unit tests only.
- Deep research source counts vary by answer; an answer that does not search returns none, which is correct rather than a failure.
- Gemini browser mode over MCP is missing the same executor dispatch this PR adds for Perplexity. That defect is pre-existing and deliberately left out of scope here.


mkubenka:
@clawsweeper re-review

Two follow-up commits since the reviewed head `edef78eb`:

- `487988b7` — the earlier review round: verify the send on follow-up turns instead of treating the conversation URL as proof; stop the follow-up resolver rebuilding a ChatGPT URL from a non-ChatGPT conversation id; check every requested model against the browser-only guard; prefer the resolved model id over the browser picker label when choosing the Perplexity mode; close the tab a run opens when Chrome is kept alive.
- `242f50ad` — this round's two findings: route every Perplexity check through one predicate that matches only the bare browser ids, so `perplexity/sonar-pro` stays on the API path (the prefix test had been copied into six places and had already drifted), and persist saved images as session artifacts so an MCP image generation reports the file it wrote.

The PR body now carries a Proof section with the gate results, live Search / Deep research / image / upload / follow-up runs, the routing guards each exercised, and an explicit limitations list.

mkubenka:
@clawsweeper re-review

`9a33d8f6` addresses the alias-routing finding. The check accepted any unqualified id starting with `perplexity`, so `perplexity-labs` was rejected on the API path and silently ran Search on the browser path. Aliases are now matched exactly from one table (`perplexity`, `perplexity-search`, `pplx`, `perplexity-research` and its spellings), an unrecognized `perplexity-*` id is kept verbatim so the browser gate rejects it by name rather than falling through to the GPT label heuristics, and `resolvePerplexityWebModel` no longer coerces arbitrary strings containing "research". Regression cases cover `perplexity-labs`, a qualified id, and the alias table.

Both behaviours are verified through the CLI: `--engine api --model perplexity-labs` passes through to the custom-model path unchanged, and `--engine browser --model perplexity-labs` is rejected by name.

The PR body now also carries a redacted trace of a signed-in Search run through the built CLI — provider log lines, answer, `## Sources`, and the run summary — plus what the MCP `consult` result contains. Account identifiers and the conversation id are omitted; nothing else is edited. The limitations list is unchanged and still states what is not covered: no live test file, `--browser-attachments always` unexercised, reattach unverified, and the Cloudflare fail-fast path unit-tested only.

steipete:
Batch-two maintainer check: the available Chrome profile is still signed out of Perplexity, so fresh signed-in validation of this revision is unavailable. Holding this PR with LAND intent until that proof path is available.

The maintainer repair remains: preserve existing bare API model IDs, track fresh image output per turn, and retain Markdown in captured answers. We will make those changes and verify Search, research, formatted answers, and image follow-ups; no contributor revision is requested.

steipete:
Fresh batch-two check through the same retained Chrome connection used for the other browser proofs: Perplexity still displays “Sign In.” No signed-in Search/research/image-follow-up validation is available from this profile, so this remains blocked with LAND intent.

The maintainer-owned repair remains preserving bare API model IDs, tracking fresh image output per turn, and retaining Markdown, followed by signed-in proof. No contributor revision is requested.

steipete:
Closing this revision under the current land-or-close maintenance pass. It still rejects previously accepted bare Perplexity API model IDs, uses conversation-wide image counts for completion, and captures answer text through `innerText` before assigning it to `answerMarkdown`. Those paths can break existing API configurations, reuse an earlier turn's images, and lose Markdown structure.

Fresh inspection of the available browser profile still shows Sign In, so the required authenticated Search/research/formatted-answer/image-follow-up proof is unavailable. The earlier LAND intent remains recorded as a recommendation to reopen once the maintainer-owned repairs and authenticated proof can be completed. No contributor revision is requested.

PR #441, opened 2026-09-01
fix(browser): verify localized effort labels with Unicode boundaries

### LINKED ISSUE #440: Browser: make direct effort slider parsing locale-independent
# Browser: direct effort slider parsing is coupled to localized announcement grammar

## Summary

The direct thinking-effort slider added in #424 exposes structured numeric state on its slider thumb, but Oracle still parses localized `aria-describedby` prose to recover the selected label.

Current main extracts the leading label with `split(/[,，]/)`. This makes the adapter depend on an allowlist of locale-specific punctuation:

- en-US: `Pro, 5 of 5`
- pt-BR: `Pro, 5 de 5`
- ja-JP: `Pro、5件中5件目。` (also reported in #438)

The pt-BR form currently works because it uses an ASCII comma. Adding another delimiter, as proposed for ja-JP in #438, fixes one locale but leaves the parser structurally fragile.

## Environment and reproduction evidence

- Oracle: `@steipete/oracle@0.18.0` / current main direct-slider implementation
- Engine: ChatGPT browser mode
- Requested model: `gpt-5.6-sol`
- Requested effort: `pro`
- UI locale: `pt-BR`

The pt-BR slider announcement is `Pro, 5 de 5`; ja-JP announces `Pro、5件中5件目。`. The direct-slider DOM contract established by #424 exposes the thumb's current position through `aria-valuemin`, `aria-valuemax`, and `aria-valuenow`.

On current main, pt-BR succeeds because the parser happens to split on its comma, while the ja-JP announcement returns `selection-unverified` because `、` is not in the delimiter allowlist. Other punctuation or a plain whitespace boundary fail for the same reason.

A narrow local verifier had added connector-specific parsing for `of|de` and completed a real pt-BR browser canary with exact `Pro` selection, verified Sol model evidence, and a persisted response. That workaround is intentionally not proposed upstream because the generic adapter should not parse positional grammar at all.

## Expected behavior

The direct-slider adapter should:

1. navigate from structured slider state rather than parsing localized ordinal prose;
2. positively verify the requested effort label after movement;
3. keep strict Pro fail-closed semantics;
4. reject contradictory or malformed evidence;
5. avoid requiring one regex branch per locale's punctuation and positional grammar.

Numeric position alone must never prove Pro.

## Proposed direction

Separate the two evidence channels:

### 1. Numeric navigation evidence

Read and validate `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` from `[role="slider"]`.

- Accept only the known five-tier range.
- Use the validated current index to calculate bounded arrow-key movement.
- Reject incomplete, non-numeric, out-of-range, or contradictory attributes.

### 2. Selected-label evidence

Read the elements referenced by the owning control's `aria-describedby`, but match the selected tier as an anchored Unicode-aware label prefix instead of splitting on locale-specific punctuation or parsing the trailing ordinal phrase.

For strict Pro, require an exact `Pro` label boundary at the beginning of one announcement. A boundary may be whitespace, a Unicode punctuation character, or end-of-string. This accepts `Pro, ...`, `Pro，...`, and `Pro、...` without accepting `Professional` or a position-only announcement.

After keyboard movement, require both:

- the structured thumb index for the requested tier; and
- the requested selected-label evidence.

If numeric and label evidence disagree, return `selection-unverified`.

The human-readable positional suffix can remain diagnostic text, but should not be a required parser input.

## Suggested tests

- `Pro, 5 of 5`
- `Pro, 5 de 5`
- `Pro、5件中5件目。`
- fullwidth punctuation
- exact `Pro` with valid thumb state
- position-only `5 of 5` rejected as Pro
- `Professional, 5 of 5` rejected
- malformed/incomplete thumb attributes rejected
- label/index contradiction rejected
- stale announcement after `aria-valuenow` changes rejected
- current English, German, Japanese, Chinese, and Korean picker tests remain green

## Related work

- #424 introduced the direct-slider adapter.
- #438 fixes the immediate ja-JP ideographic-comma failure.
- #423 demonstrates the ongoing cost of locale-specific picker vocabulary.

This issue proposes a follow-up that removes announcement-delimiter coupling while preserving exact tier verification.

Gabrielgvl:
The direct effort slider announces the selected tier using localized text, such as Japanese `Pro、5件中5件目。`. Main splits that announcement only at ASCII/fullwidth commas, so it cannot verify this Japanese form or navigate through Japanese announcements. Match a known label at the start of the announcement, independently verify the five-tier numeric ARIA position, and allow whitespace, Unicode punctuation, or end-of-string after the label.

The boundary check must inspect the original Unicode suffix. Normalizing a single UTF-16 code unit erases characters such as `é`, Cyrillic letters, astral digits, combining marks and symbols; that incorrectly verifies `Proé` as Pro. This candidate rejects those continuations and retains the existing fail-closed behavior for missing labels, stale/contradictory state, invalid ranges and position-only announcements. The false positives also reproduce on main; they are not attributed solely to the initial PR.

This preserves @Gabrielgvl's implementation and commit history, incorporates @kiyo-e's Japanese reproduction, and merges current main without rewriting contributor commits. Thanks to both contributors. The narrower #438 overlaps with the Japanese coverage here and remains open pending maintainer disposition.

## Real browser proof

A local synthetic page exercised the production selection expression in the existing Chrome browser through the allowed Chrome extension. The corrected expression came from the built package. No DOM stubs, timer substitutions, provider requests, signed-in ChatGPT changes, or isolated-browser substitution were used for this proof.

| Candidate | Expected outcomes |
| --- | --- |
| Current main | 3 / 11 |
| Initial PR #441 | 6 / 11 |
| Corrected candidate | 11 / 11 |

The matrix verifies Japanese already-selected Pro and two ArrowRight movements from High through Extra High to Pro with Japanese announcements, Portuguese and whitespace forms, Unicode continuation rejection, numeric contradiction, and position-only rejection. This proves the changed locale parsing and keyboard behavior on the documented synthetic DOM shape; it does not claim a new live ChatGPT consultation.

Before (main):

Main rejects Japanese announcements and accepts malformed Unicode continuations

After:

Corrected built parser passes all eleven synthetic real-Chrome cases

## Validation

- Six added Unicode regression cases failed against the original PR and passed after correction.
- `pnpm test`: 154 files / 1,938 tests passed, 18 files / 43 tests skipped. Existing stale-state, malformed-range, non-Pro and multilingual tests remain green.
- `pnpm run check`, `pnpm run build`, `pnpm run docs:check`, `pnpm run docs:site`, and `pnpm run test:packed-cli` passed.
- Full candidate P0–P2 Codex review: scoped-clean, no actionable findings.

Closes #440.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Browser: recognize collision-renamed attachment chips, including short filenames, while keeping Unicode filename boundaries and visible extensions distinct across upload and send checks. Fixes #393. Thanks @devYRPauli!
 - Browser: select and verify thinking effort in ChatGPT's direct-slider picker without an Advanced submenu, keeping explicit Pro requests fail-closed. Fixes #422.
 - Browser: preserve Hangul and recognize Korean model/effort picker labels, distinguishing High from Extra High and keeping Pro verification strict. Fixes #423.
+- Browser: verify direct-slider effort labels without depending on locale-specific announcement punctuation or ordinal grammar. Fixes #440.
 - Browser: restore locally launched macOS Chrome windows to their prior placement for visible runs only when Oracle recorded the window before a `--browser-hide-window` run; unmarked windows remain untouched.
 - Browser: archive completed ChatGPT conversations when the interface is Japanese by recognizing the current `その他`, `アーカイブ`, and `アーカイブを解除する` controls.
 - Browser: apply the configured input timeout to prompt preparation so stalled local file assembly fails clearly before launching Chrome. Fixes #381.
--- a/src/browser/actions/thinkingTime.ts
+++ b/src/browser/actions/thinkingTime.ts
@@ -1064,17 +1064,38 @@ function buildThinkingTimeExpression(
         const thumb = slider?.querySelector?.('[role="slider"]');
         if (!control || !thumb || !isVisible(control)) return null;
         const levels = ['light', 'standard', 'extended', 'extra-high', 'pro'];
-        const labels = describedIds(control)
-          .map((id) => (document.getElementById?.(id)?.textContent ?? '').split(/[,，]/)[0].trim())
-          .filter((label) => levels.some((level) => TARGET_LEVEL_TOKENS[level].some((token) => normalize(token) === normalize(label))));
-        if (labels.length !== 1) return null;
-        const label = labels[0];
-        const index = levels.findIndex((level) => TARGET_LEVEL_TOKENS[level].some((token) => normalize(token) === normalize(label)));
+        // The selected label leads localized aria-describedby prose, while punctuation
+        // and ordinal grammar vary by locale. Match only that leading label and let
+        // the thumb's numeric ARIA state independently prove its position.
+        const readLeadingLevel = (description) => {
+          const value = (description ?? '').normalize('NFC').trim();
+          if (!value) return null;
+          const matches = levels.flatMap((level, index) => {
+            const token = [...TARGET_LEVEL_TOKENS[level]]
+              .sort((left, right) => right.length - left.length)
+              .find((candidate) => {
+                const normalizedCandidate = candidate.normalize('NFC');
+                const prefix = value.slice(0, normalizedCandidate.length);
+                if (normalize(prefix) !== normalize(normalizedCandidate)) return false;
+                const next = normalize(value.slice(prefix.length, prefix.length + 1));
+                return !next;
+              });
+            return token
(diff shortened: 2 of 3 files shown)


Gabrielgvl:
Added redacted real-browser proof to the PR body from an after-fix pt-BR ChatGPT browser run. The deployed artifact was inspected for the generic matcher and absence of both the direct-digit boundary and `of|de` connector parser before the canary; the run then verified GPT-5.6 Sol + Pro and returned the exact synthetic marker without API fallback.

@clawsweeper re-review

Issue #435, opened 2026-08-29
Browser mode: recovered/error runs can release the lease while retaining Oracle-owned tabs indefinitely


lhysin:
## Summary

In browser mode with a persistent manual-login profile, an Oracle-owned ChatGPT tab can survive an error after its browser lease is released. If the tab is intentionally kept for recovery, a later successful `session --harvest` / reattach still appears to close only the CDP attachment, not retire the original Oracle-owned page target.

This means retained error/recovery tabs can accumulate independently of `maxConcurrentTabs`: the active lease count stays bounded, but old Oracle-owned targets can remain in the shared Chrome while later runs acquire freed leases and open new targets.

I am **not** suggesting that all failed/rate-limited tabs should be closed immediately. I have local cases where a run first failed on a visible ChatGPT rate-limit warning and a later harvest recovered useful output. The gap is the lack of a bounded retirement path after recovery is no longer needed or after recovery succeeds.

## Environment

```text
Oracle CLI: 0.18.0
Engine: browser
Profile: persistent manual-login profile
maxConcurrentTabs: 3
Platform: macOS
```

I also checked current `main` at `bbc1b3b0261d` (2026-08-28); the lifecycle described below is still present there.

## What I observed

During repeated browser work across multiple independent work units, the shared Oracle Chrome accumulated dozens of old ChatGPT tabs while the configured active lease limit remained bounded.

The persistent Chrome profile also retained session state containing roughly 48-49 distinct ChatGPT conversation URLs in its recent `Default/Sessions/Tabs_*` files. I do not claim every retained conversation came from this exact path, but it is consistent with the visible tab accumulation.

For the rate-limit trigger specifically, I found eight local Oracle sessions that correctly recorded the ChatGPT modal as:

```text
code: chatgpt-ui-warning
uiWarning.type: rate_limit
message: Too many requests ... You're making requests too quickly ...
```

At least three of those sessions were later harvested successfully (`harvest.state=completed`), which is why immediate close-on-error would be unsafe.

## Source-level lifecycle

Current `src/browser/index.ts` has:

```ts
function shouldCloseOwnedRunTargetAfterRun(options: {
  runStatus: "attempted" | "complete";
  ownsTarget: boolean;
  keepBrowser: boolean;
  closeOwnedTabOnComplete?: boolean;
}): boolean {
  return (
    options.runStatus === "complete" &&
    options.ownsTarget &&
    (Boolean(options.closeOwnedTabOnComplete) || !options.keepBrowser)
[...]

Issue #390, opened 2026-08-16
Harvest can capture another concurrent run's answer and report it as this run's completed result


postoso:
**Version:** oracle 0.17.3, browser engine (`-e browser --browser-manual-login`), persistent Chrome profile, GPT-5.6 Sol at Pro thinking time

When two runs are launched against the same persistent Chrome profile close together, a run can bind to another run's conversation. The harvest then captures **that other run's answer** and stores it under this run's slug. The session reports `completed`, the archived transcript passes validation, and nothing in the output indicates the answer belongs to a different question.

## What I observed

Session slug `oracle-trigger-rewrite-contract-review`. Its archived transcript at `~/.oracle/sessions/<slug>/artifacts/transcript.md` contains:

- `## Prompt` — a request to review an operating-reference document for a browser-automation lane. Correct; this is the prompt that was sent.
- `## Answer` — a twenty-section review of an entirely different subject: a database schema-recovery runbook, discussing Dolt commits, `schema_migrations` rows, and backup semantics. It has no relationship to the prompt above it.

A second run against the same profile that evening was reviewing exactly that database runbook. Its answer is what landed in this session's transcript.

The run's own header reported:

```
Status: completed
Response: status=completed
Artifacts:
- Browser transcript — .../artifacts/transcript.md (36.5 KB) sha256=7ccd3dd9d23b… validation=ok
```

So `validation=ok` and `sha256` are present and correct. They validate that the file transferred intact, not that its contents answer the prompt. Both signals a caller would reach for to check success are green.

## Why this is worth a guard

Every other failure mode in this area is detectable by counting or by an error. This one is not. The output is a well-formed, high-quality answer to *a* question, so it survives a skim by a human and passes any automated check that looks at status, exit code, size, or hash. Where a caller is chaining oracle output into further automated work, a topically adjacent answer can be acted on before anyone notices.

## Suggested direction

I do not have a deterministic reproduction; it requires two runs racing on one profile and I hit it once. But the failure seems addressable without reproducing the race:

- **Bind and verify a conversation id per run.** If the harvest is scoped to a conversation id the run itself established, capturing a different conversation's content becomes detectable rather than silent.
- **Treat a mismatch as an error, not a completion.** Right now the run reports `completed`.
- **A caller-side marker also works and needs no server changes.** I now open every prompt with an instructed echo token and refuse to trust a harvested answer whose first line does not carry it. That has been effective, but it is a workaround each caller has to invent, and it only helps callers who know to do it.

[...]

PR #445, opened 2026-09-02
fix(browser): preserve shared Chrome across concurrent sessions

oraclexing:
Preserve shared manual-login Chrome while peer Oracle controllers still own leases, and let the verified final lease owner perform cleanup under the registry lock. Native Windows shared-profile Chrome is detached from its launching controller; temporary and copied profiles retain their existing lifecycle.

Maintainer repair addresses five reproduced failures in the submitted head: replacement of a held empty POSIX lock, reaping a malformed owner as ownerless, live-owner recovery after false native PID probes, and stale-lease loss despite OS-reported process start identity (including for the current controller) or recovered liveness. Acquisition now uses exclusive mkdir; malformed owners are retained; OS-reported process start identity (including for the current controller) takes precedence over a transient PID miss, with repeated negative observations required when identity is unavailable. Owner records are atomically published so interrupted initialization is recoverable without deleting malformed ownership evidence; regression tests cover partial writes and abandoned temporary owners. Registry writes and failed lock initialization clean up only what they can verify they own.

The repair also rejects invalid lease PIDs and ambiguous process/profile command lines, supports a quoted complete Windows profile argument, handles POSIX process-list profile paths containing spaces and standard Chrome Beta/Dev/Testing channel names, unrefs detached Windows children so the launching controller can exit, and bounds process queries. All controllers should be restarted together after upgrading: stored legacy records remain readable, but older live controllers can forcibly remove locks after their own timeout.

Validation: 2,106 full-suite tests passed, 45 skipped; typecheck/lint/build/docs passed. The five new safety regressions failed before the repair. The current built-CLI macOS proof starts two controllers with one fresh shared Chrome, verifies two distinct targets and one send each, observes the owner exit while the peer passes 20 CDP/Runtime checks, then verifies the peer's final release, empty registry, process/endpoint shutdown, and distinct persisted answers. Linux and native Windows run the same proof in CI; exact-head results and final autoreview are recorded in the proof comment.

The test's initial homepage requests are supplied locally through CDP interception in an isolated, signed-out Chrome; no ChatGPT request or authentication challenge is bypassed. This does not establish fresh signed-in ChatGPT/Deep Research concurrency, backend model or effort identity, or WSL behavior. Signal and temporary/copied-profile policies have focused regression coverage; the full CLI scenario covers normal controller completion/exit.

Thanks @oraclexing. Release-note context for the maintainer after landing: Browser: preserve shared Chrome across concurrent manual-login controllers, verify final lease ownership before shutdown, and harden lock recovery against transient process probes; thanks @oraclexing. No CHANGELOG.md change is included here, keeping #462 limited to its specified seven-change set.


oraclexing:
@clawsweeper re-review

Author correction for head `728f03df`:

The PR body's initial causal attribution is superseded. Multi-process registry stress did not reproduce lease loss. The higher-confidence Windows root cause was controller process ownership: `chrome-launcher` started the shared Chrome as a normal child of the controller that launched it, so controller exit could close Chrome underneath still-valid peer leases.

The new commit adds native-Windows `detached: true` + `windowsHide: true`, preserves non-Windows/WSL behavior, drains in-flight lease heartbeats before release, adds exact final/non-final telemetry, and makes both launched and reused manual-login final cleanup use the verified PID/profile terminator (hidden tree kill + post-kill liveness polling). It also fixes the POSIX compatibility finding by making Windows path/case assertions Windows-scoped.

Exact-head verification:

- lint/typecheck, build, and diff-check passed
- focused browser suite: 5 files, 140 passed, 1 skipped
- heartbeat/release overlap test passed 10 consecutive runs
- full suite with 8 workers: 156 files passed / 18 skipped; 1,977 tests passed / 53 skipped
- independent lifecycle review: GO, no P0/P1

Real Windows two-controller smoke:

- owner launched Chrome PID 48300 / port 57689; peer acquired another lease and target
- owner release was non-final with peer still recorded, then owner controller exited
- after owner exit, peer completed 20 consecutive `/json/version`, `/json/list`, and target-scoped `Runtime.evaluate` checks
- peer release was final; verified termination succeeded; registry empty and endpoint unreachable

Final-kill smoke separately verified PID 14808 / port 64198 alive+reachable before cleanup and both absent after; no new WindowsTerminal/OpenConsole process appeared.

Evidence boundary: a later Deep Research session was manually submitted by the user, so its final `promptSubmitted=true` must not be treated as automatic-send proof. Exact-head headless ChatGPT attempts were blocked by Cloudflare; this head claims exact process/lease/CDP lifecycle proof, not a new signed-in automatic-submit proof.

steipete:
Maintainer repair verified on `151912d296798759f43155b13f15440f16999042`. All four CI jobs passed, including the real native Windows and Linux shared-Chrome lifecycle proof: https://github.com/steipete/oracle/actions/runs/34123633188.

The submitted Windows controller-ownership fix is retained and scoped to shared manual-login profiles. Detached children are unreferenced so their launching controller can exit. Exclusive directory creation prevents replacement of a live empty POSIX lock; OS-reported process start identity (including for the current controller) corroborates liveness; malformed owners remain preserved; atomically published owner records prevent interrupted initialization from stranding partial ownership. Exact process/profile matching supports spaces and standard Chrome channel names while rejecting ambiguous flags. All controllers must restart together after upgrade because older live versions have unsafe lock-timeout recovery.

Validation: 2,106 full-suite tests passed, 45 skipped; typecheck/lint/build/docs passed. Five safety regressions failed on the submitted code before repair; new interrupted-write, abandoned-temporary-owner, and wall-clock-correction regressions pass. Local and final branch autoreview through P2 are scoped-clean.

The current built CLI and real Chrome passed on macOS, Linux, and native Windows with a profile path containing spaces: two distinct targets, one send per controller, normal owner exit with the peer lease retained, twenty successful peer CDP/version/target/runtime checks afterward, peer completion and final release, empty registry, absent lock, dead Chrome PID, unreachable endpoint, and distinct persisted answers. Windows console-process snapshots found no additional WindowsTerminal/OpenConsole processes at the end. Historical Windows failure evidence remains the contributor's corrected trace; this is fresh forward proof on the repaired head.

Evidence boundary: the isolated signed-out profile receives synthetic homepage/application responses locally; this does not prove signed-in ChatGPT or Deep Research concurrency, actual backend model/effort, WSL, or long-duration production behavior. Beta/Dev/Testing executable-name handling has regression coverage; those channel binaries were not separately launched. The real CLI scenario uses normal completion and exit; abrupt-signal and temporary/copied-profile policies have focused unit coverage only. #390 and #435 are not claimed resolved.

Thanks @oraclexing. Human coauthor credit is preserved. Release-note credit remains in the PR body for the maintainer after landing; no merge or release was performed.

PR #452, opened 2026-09-05
fix(browser): stop remote runs when isolated tab creation fails

ShunmeiCho:
Ordinary remote browser runs now stop if their dedicated tab cannot be created or attached. Previously, the connection helper could fall back to a default target, leaving subsequent navigation pointed at an unrelated conversation.

Add opt-in `fallbackToDefault: false` and enable it for ordinary remote runs. Strict calls without a target URL create `about:blank`; callers that omit the option retain their existing helper behavior. Cleanup remains scoped to newly created targets.

## Real Chrome evidence

[Protocol and scope](https://github.com/ShunmeiCho/oracle/blob/aaf9fe90de0b34dcc8e07599dc00e269cf28fd6b/docs/evidence/pr-review/README.md) · [candidate results](https://github.com/ShunmeiCho/oracle/blob/aaf9fe90de0b34dcc8e07599dc00e269cf28fd6b/docs/evidence/pr-review/452/candidate.json) · [baseline results](https://github.com/ShunmeiCho/oracle/blob/aaf9fe90de0b34dcc8e07599dc00e269cf28fd6b/docs/evidence/pr-review/452/baseline.json) · [reproduction script](https://github.com/ShunmeiCho/oracle/blob/aaf9fe90de0b34dcc8e07599dc00e269cf28fd6b/scripts/remote-tab-isolation-proof.mjs)

This uses an independent temporary Chrome 137/Linux profile and real HTTP/WebSocket transport. A loopback proxy injects target-creation HTTP 500 or a new target's WS-handshake HTTP 502; it does not mock the CDP client. The matrix covers a fresh profile and the same running browser after another protected page exists.

| Scenario, in both setups | Baseline `be6c92a` | Candidate `7f8e619` |
| --- | --- | --- |
| Creation failure | Connects to protected sentinel | Rejects; zero sentinel WS connections |
| Attachment failure | Connects to protected sentinel | Rejects; closes only newly created target |
| Normal connection/cleanup | Dedicated target works | Dedicated target works; only its own target closes |

Candidate passes all six cases. The identical strict assertions fail on the four baseline fallback cases. Records include module hashes, actual WS target counts, unchanged sentinel URL/nonce/DOM hashes, zero sentinel navigations, and restored target sets. Success also executes and verifies a title update on the new target. Cleanup completes before the final success marker. No ChatGPT login or model request is involved.

## Compatibility and validation

[...]

PR #454, opened 2026-09-05
fix(browser): block sends while attachments are uploading

### LINKED ISSUE #446: Browser: attachment completion can resolve while uploading=true, permitting an early send path
## Summary

On current `main` (`40c188caa861017ba48e7b496e2171317c648976`), `waitForAttachmentCompletion()` can return when its production DOM probe reports `state="ready"`, `uploading=true`, and valid per-file evidence for every expected attachment.

Once all expected files have per-file evidence and Send is ready, `uploading=true` only increases the stability window from 1.5 s to 3 s. After that window, the completion gate returns. The final Send gate rechecks attachment identity and button readiness, but does not recheck upload status.

If that `uploading=true` signal corresponds to a later attachment that is registered but not finished, this leaves a premature-send path: the completion gate can return after 3 s, and the next gate can click Send without reading upload status again.

The reproduction below executes the production completion expression against the repository's controlled DOM fixture, including two retained per-file receipts, an enabled Send button, and an in-composer upload-progress node. It is not a signed-in ChatGPT end-to-end reproduction on current `main`, and it does not prove that `uploading=true` always corresponds to an unfinished network transfer. The current upload-status query is document-wide, so live frequency and meaning depend on ChatGPT's DOM.

The motivating live UI symptom was observed on an older pinned Oracle commit: during a multi-file submission, Send became active after one file and Oracle submitted before the remaining attachment appeared to finish. Bundling the same inputs into one ZIP succeeded end to end. That observation motivates this check but does not establish current-main live reachability.

## Minimal reproduction

Add this test as `tests/browser/attachmentUploadRace.test.ts`:

```ts
import { expect, test, vi } from "vitest";
import { buildAttachmentEvidenceExpression } from "../../src/browser/actions/attachmentEvidence.js";
import { waitForAttachmentCompletion } from "../../src/browser/pageActions.js";
import type { ChromeClient } from "../../src/browser/types.js";
import { FakeDocument, FakeElement, FakeInputElement } from "./domFixture.js";

test("does not finish while a multi-file composer still reports uploading", async () => {
  vi.useFakeTimers();
  vi.setSystemTime(0);
  try {
    const form = new FakeElement("form", { "data-testid": "composer" }, [
      new FakeElement("textarea", { id: "prompt-textarea" }),
      new FakeInputElement([]),
      new FakeElement("div", { "data-testid": "upload-progress", "data-state": "uploading" }),
      new FakeElement("button", { "data-testid": "send-button" }),
    ]);
    const document = new FakeDocument([form]);
    const renderer = {};
    const evaluate = (expression: string) =>
      new Function(
        "document", "HTMLElement", "HTMLInputElement", "window", "globalThis",
        `return ${expression};`,
      )(
        document,
        FakeElement,
        FakeInputElement,
        { getComputedStyle: () => ({ pointerEvents: "auto" }) },
        renderer,
      );
    const runtime = {
      evaluate: vi.fn(async ({ expression }: { expression: string }) => ({
        result: { value: evaluate(expression) },
      })),
    } as unknown as ChromeClient["Runtime"];

    for (const [id, name] of [["first", "a.txt"], ["second", "b.txt"]]) {
      evaluate(buildAttachmentEvidenceExpression([name], "begin", id));
      const remove = new FakeElement("button", { "aria-label": `Remove file: ${name}` });
      form.append(new FakeElement("div", { "data-testid": "attachment-chip" }, [remove]));
      expect(evaluate(buildAttachmentEvidenceExpression([], "confirm", id))).toBe(true);
    }

    const completion = waitForAttachmentCompletion(runtime, 5_000, ["a.txt", "b.txt"]);
    const outcome = completion.then(
      () => "resolved",
      () => "rejected",
    );
    await vi.advanceTimersByTimeAsync(6_000);
    expect(await outcome).toBe("rejected");
  } finally {
    vi.useRealTimers();
  }
});
```

Run:

```bash
pnpm exec vitest run tests/browser/attachmentUploadRace.test.ts --reporter verbose
```

Actual result:

```text
AssertionError: expected 'resolved' to be 'rejected'
Expected: "rejected"
Received: "resolved"
```

## Root cause

1. In [`waitForAttachmentCompletion()`](https://github.com/steipete/oracle/blob/40c188caa861017ba48e7b496e2171317c648976/src/browser/actions/attachments.ts#L1661-L1675), `uploading=true` changes only `stableThresholdMs`; the ready return does not require `uploading === false`.
2. The [file-input fallback](https://github.com/steipete/oracle/blob/40c188caa861017ba48e7b496e2171317c648976/src/browser/actions/attachments.ts#L1707-L1727) has the same property.
3. The final [`attemptSendButton()` attachment gate](https://github.com/steipete/oracle/blob/40c188caa861017ba48e7b496e2171317c648976/src/browser/actions/promptComposer.ts#L515-L540) verifies chips/receipts/file inputs, and the [click path](https://github.com/steipete/oracle/blob/40c188caa861017ba48e7b496e2171317c648976/src/browser/actions/promptComposer.ts#L611-L651) verifies an enabled, stable button, but neither reads the upload state.
4. The existing `uploading=true` input-only test uses an 800 ms timeout, so it never crosses the 3 s stability window that triggers this path.

Per-file evidence proves that each assignment owns a still-connected removal control in the composer. It is not itself evidence that the corresponding network upload completed.

## Expected behavior

A trustworthy, composer-scoped in-progress signal should prevent attachment completion and Send. If the current document-wide `uploading` probe is too broad or unreliable for that role, it should be scoped or replaced with stronger completion evidence. In either design, an explicitly true in-progress signal should not become "complete" merely because three seconds elapsed.

## Actual behavior

In the controlled DOM fixture, two files have valid receipts, Send is enabled, and an in-composer node matches the production upload-status selectors. The completion gate still resolves after roughly 3 s. Source inspection shows that the subsequent Send gate does not query upload status.

## Diagnostic fix check

As a diagnostic only, adding `value.uploading === false` to the all-files-matched ready return made the new reproduction pass. All 57 existing tests in `tests/browser/attachmentsCompletion.test.ts` also passed. I reverted that change afterward; a complete fix should audit every return path and the final Send gate rather than rely on this one-line probe alone.

## Relation to prior work

- #434 retained per-file evidence across upload, completion, and send readiness and stabilized the one-click send path. This report is about a remaining completion-state invariant after that identity fix.
- #418 covered an attachment-bearing send that silently no-opped.
- #115 covered a timeout path that continued into an Enter fallback.

This is a different remaining completion invariant after #434's fix for #418. #115 concerned the timeout-to-Enter fallback rather than successful completion while the upload probe remains true.

steipete:
Attachment completion could succeed after three seconds while the DOM still reported an upload in progress. The final send gate checked attachment identity but did not recheck progress, permitting an early send.

Use one composer-scoped structured-state detector at completion and final send readiness. Explicit loading/busy states and incomplete native/ARIA progress controls block sending and reset stability windows. Busy state on hidden file inputs or backing editors remains authoritative. Respect the nearest editing boundary so non-editable attachment widgets nested inside a rich-text editor retain their upload state, while prompt-owned editable markup is excluded. Filenames and status prose are not used to infer transfer state.

Fixes #446; thanks @HJC704.

Independently rebased onto main at 748193758d54e43779ab93e94fac23cbc6c9e421. Current head: 4f207cb0ad5e8f350759005bf7599b2075b91282. This PR has no CHANGELOG.md diff; #455 owns its release-note coverage and contributor thanks. The restack preserved the existing source; subsequent autoreview exposed the non-editable-widget edge case, reproduced in three new tests and in the built real-Chrome proof before correction.

Fresh validation on Node 24.20.0/macOS: 179 attachment/composer tests passed; typecheck, lint, formatting, build and built CLI docs/help checks passed. The updated real-Chrome proof (`node scripts/attachment-send-proof.mjs` after building) passed for local and remote three-file uploads, filename-less image evidence, active/completed SVG/native progress, opacity and offscreen recovery. Holding upload state active inside a non-editable rich-editor widget for 4.5 seconds blocks completion; attempted send produces zero clicks/Enter events. Clearing state permits exactly one trusted click and one committed turn with byte-identical attachments. This is controlled browser-runtime proof, not a claim about current signed-in ChatGPT race frequency.

The staged candidate passed autoreview through P2. Exact-head branch review and CI results are recorded in the final proof comment. No merge performed.

### DIFF
--- a/docs/manual-tests.md
+++ b/docs/manual-tests.md
@@ -22,7 +22,7 @@ Run `pnpm build && node scripts/attachment-send-proof.mjs` with Chrome installed
 
 The disposable profile and fixtures live in a temporary, non-hidden directory under your home directory and are removed afterward. This lets Snap Chromium read the same files as Node instead of looking in its private `/tmp`. An optional directory argument retains the fixtures for manual testing; that directory must also be readable by the selected browser.
 
-The attachment proof also holds composer upload progress active for longer than three seconds, verifies completion and send both refuse it without input, then clears it and verifies one successful send despite unrelated page progress. Readiness uses explicit loading/busy state and native/ARIA progress controls; filenames and status prose alone cannot establish an active transfer.
+The attachment proof also holds composer upload progress active for longer than three seconds inside a non-editable attachment widget nested in a rich-text editor, verifies completion and send both refuse it without input, then clears it and verifies one successful send despite unrelated page progress. Readiness uses explicit loading/busy state and native/ARIA progress controls; filenames and status prose alone cannot establish an active transfer.
 
 ### Quick browser port smoke
 
--- a/docs/manual-tests.md
+++ b/docs/manual-tests.md
@@ -22,6 +22,8 @@ Run `pnpm build && node scripts/attachment-send-proof.mjs` with Chrome installed
 
 The disposable profile and fixtures live in a temporary, non-hidden directory under your home directory and are removed afterward. This lets Snap Chromium read the same files as Node instead of looking in its private `/tmp`. An optional directory argument retains the fixtures for manual testing; that directory must also be readable by the selected browser.
 
+The attachment proof also holds composer upload progress active for longer than three seconds, verifies completion and send both refuse it without input, then clears it and verifies one successful send despite unrelated page progress.
+
 ### Quick browser port smoke
 
 - `pnpm test:browser` — launches headful Chrome and checks the DevTools endpoint is reachable. Set `ORACLE_BROWSER_PORT` (or `ORACLE_BROWSER_DEBUG_PORT`) to reuse a fixed port when you’ve already opened a firewall rule.
--- a/docs/manual-tests.md
+++ b/docs/manual-tests.md
@@ -22,7 +22,7 @@ Run `pnpm build && node scripts/attachment-send-proof.mjs` with Chrome installed
 
 The disposable profile and fixtures live in a temporary, non-hidden directory under your home directory and are removed afterward. This lets Snap Chromium read the same files as Node instead of looking in its private `/tmp`. An optional directory argument retains the fixtures for manual testing; that directory must also be readable by the selected browser.
 
-The attachment proof also holds composer upload progress active for longer than three seconds, verifies completion and send both refuse it without input, then clears it and verifies one successful send despite unrelated page progress.
+The attachment proof also holds composer upload progress active for longer than three seconds, verifies completion and send both refuse it without input, then clears it and verifies one successful send despite unrelated page progress. Readiness uses explicit loading/busy state and native/ARIA progress controls; filenames and status prose alone cannot establish an active transfer.
 
 ### Quick browser port smoke
 
--- a/scripts/attachment-send-proof.mjs
+++ b/scripts/attachment-send-proof.mjs
@@ -175,6 +175,30 @@ try {
       /Attachments never reached a clickable send button/,
     );
     assert.equal(await evaluate("window.proof.clicks + window.proof.enters"), 0);
+    await evaluate('document.querySelector("#upload-progress").style.opacity = "0"');
+    assert.equal(await evaluate(buildAttachmentReadyExpressionForTest(names)), true);
+    await evaluate('document.querySelector("#upload-progress").style.opacity = "1"');
+    assert.equal(await evaluate(buildAttachmentReadyExpressionForTest(names)), false);
+    await evaluate(`(() => {
+      document.querySelector('#upload-progress').remove();
+      const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+      svg.id = 'upload-progress'; svg.setAttribute('role', 'progressbar');
+      svg.setAttribute('aria-valuenow', '50'); svg.setAttribute('width', '20'); svg.setAttribute('height', '20');
+      document.querySelector('form').append(svg);
+    })()`);
+    assert.equal(await evaluate(buildAttachmentReadyExpressionForTest(names)), false);
+    await evaluate(
+      'document.querySelector("#upload-progress").setAttribute("aria-valuenow", "100")',
+    );
+    assert.equal(await evaluate(buildAttachmentReadyExpressionForTest(names)), true);
+    await evaluate(`(() => {
+      document.querySelector('#upload-progress').remove();
+      const progress = document.createElement('progress'); progress.id = 'upload-progress';
+      progress.value = 50; progress.max = 100; document.querySelector('form').append(progress);
+    })()`);
+    assert.equal(await evaluate(buildAttachmentReadyExpressionForTest(names)), false);
+    await evaluate('document.querySelector("#upload-progress").value = 100');
+    assert.equal(await evaluate(buildAttachmentReadyExpressionForTest(names)), true);
     await evaluate(`(() => {
       document.querySelector('#upload-progress').remove();
       const unrelated = document.createElement('div'); unrelated.dataset.state = 'uploading';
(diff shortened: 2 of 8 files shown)


steipete:
Final local verification at 1b3cd7b68232cfc5a824c5c7301d3b559d638af7.

The regression is reproduced on main by extending the existing active-upload test past three seconds. The repaired completion loop now treats active progress as blocking instead of merely extending a timer, and final send readiness uses the same composer-scoped detector.

Validation: 1,965 tests passed / 43 skipped; typecheck, lint, targeted formatting, build, and built CLI docs/help check passed. The committed real-Chrome proof (node scripts/attachment-send-proof.mjs after building) passed for local and remote three-file uploads. Holding active progress for 4.5s blocks completion; an attempted send produces zero clicks and zero Enter events. Hidden/completed progress then permits one trusted click and one committed turn with exact attachment bytes. The proof also checks SVG/native progress, faded indicators, and offscreen recovery. This is controlled browser-runtime proof, not a claim of signed-in ChatGPT reproduction frequency.

Review corrections cover status-versus-filename text, terminal announcements, punctuation, opacity including ancestors, HTML/SVG progress, and completed ARIA values. The final staged candidate is autoreview scoped-clean through P2. Exact committed-branch review and fresh CI are running; their final results will follow before landing.

Fixes #446. No merge performed.

steipete:
LAND-READY at 5616637c9ea459e566917add8c1d0b786e141829.

Exact-head CI succeeded: https://github.com/steipete/oracle/actions/runs/33994036605 (Linux, macOS, Windows, Chromium proof). Final branch autoreview against origin/main is scoped-clean through P2. Full local suite: 1,977 passed / 43 skipped; typecheck, lint, formatting, build, and built CLI docs/help checks passed.

The final implementation uses explicit loading/busy state and native/ARIA progress values rather than parsing status prose. It honors busy state on hidden upload/editor controls, excludes editable content, resets completion stability while uploading, and reuses the same predicate immediately before send. The test DOM now correctly models CSS case-insensitive equality and role-token matching.

Real Chrome proof: node scripts/attachment-send-proof.mjs after building. Local and remote three-file uploads both refuse completion during 4.5s of active progress and refuse sending with zero clicks/Enter events. Clearing state permits exactly one trusted click and one committed turn with byte-identical attachments. SVG/native progress, faded indicators, filename-less images, and offscreen recovery pass. This is controlled runtime proof; no claim of a current signed-in ChatGPT race frequency is made.

Fixes #446; thanks @HJC704. No merge performed.

steipete:
LAND-READY at 4f207cb0ad5e8f350759005bf7599b2075b91282, independently based on main 748193758d54e43779ab93e94fac23cbc6c9e421.

Exact-head CI succeeded on Linux, macOS, Windows and the Chromium proof job: https://github.com/steipete/oracle/actions/runs/33995416368. Final committed-branch autoreview against origin/main is scoped-clean through P2. Fresh local validation on Node 24.20.0/macOS: 179 attachment/composer tests, typecheck, lint, formatting, build and built CLI help/docs checks passed.

The rebase itself changed only CHANGELOG.md. Autoreview then identified a real edge case: the old editable-ancestor check discarded progress inside contenteditable=false attachment widgets nested within rich-text editors. Three new regression cases failed before the correction, and the built real-Chrome proof reproduced premature completion. The final code respects the nearest valid editing boundary, retains hidden input state, and still excludes markup in nested editable hosts.

The updated committed proof ran against the built code: node scripts/attachment-send-proof.mjs. Local and remote three-file uploads held active widget progress for 4.5 seconds and refused completion; attempted send produced zero clicks/Enter events. Clearing state permitted exactly one trusted click and one committed turn with byte-identical attachments. SVG/native progress, opacity, filename-less images and offscreen recovery also passed. This is controlled browser-runtime proof, not a claim about current signed-in ChatGPT race frequency.

Fixes #446; thanks @HJC704. This PR has no CHANGELOG.md diff; #455 owns the complete release-note coverage and should land after #452 and #454. No merge performed.

PR #458, opened 2026-09-06
fix(browser): fail fast when a new turn offers Retry

developerisnow:
Fixes #457.

When the submitted ChatGPT turn displays a known English generation failure with a visible Retry control, Oracle returns a structured, body-free browser warning promptly. It preserves manual recovery and never clicks Retry or submits the prompt again. Unknown/localized failure states retain the configured timeout.

The maintainer repair keeps waiting while generation is active, ignores transparent controls, uses the outer conversation turn for busy/progress evidence, stops the pending response observer on failure, and applies the error guard to image output and response-recovery paths. The changelog entry is collected in the final maintenance-notes PR to keep sibling branches independent.

Validation:

- Typecheck, lint, formatting, docs/help consistency, and build pass.
- Complete suite: 1,999 passed, 43 skipped with four workers. An earlier default-concurrency MCP startup timeout passed when isolated and in the complete four-worker run.
- Real Chrome: `node scripts/assistant-retry-proof.mjs`, opened in an existing Chrome profile, passed nine controlled DOM cases. The production reader/error guard emitted `chatgpt-ui-warning`; older turns, hidden/transparent Retry, unrelated prose, and active generation did not fail. Zero Retry clicks.
- Built CLI: render/preview and a real OpenAI request completed successfully.
- Independent autoreview through P2 is clean for the final candidate.

The Chrome fixtures are synthetic and do not claim a newly observed live ChatGPT outage. A signed-in Oracle CDP smoke remains blocked by Chrome's separate remote-debugging approval; this is also an outstanding release gate.

Full built-CLI/CDP proof now passes for text and image failure, including persisted recovery metadata and retained Chrome: https://github.com/steipete/oracle/pull/458#issuecomment-5568090824. The separate signed-in Pro release smoke remains outstanding.

PR #430, opened 2026-08-28
feat(mcp): add detached consults and non-cancelling waits

oraclexing:
Fixes #429.

Add `consult.waitForCompletion:false` for durable local API/browser consultations and a `wait` tool that observes the saved session. Synchronous consult remains the default. Wait timeout, cancellation, and MCP transport closure stop only that waiter; remote browser-service calls remain synchronous.

The shared launcher persists the worker lifecycle before opening its start gate. The maintainer repair integrates current main, handles a closed start pipe without crashing the MCP host, and keeps changelog changes in the final notes PR. Metadata replacement retries remain bounded for transient Windows locks.

Validation:

- Build, typecheck, lint, and complete suite: 2,014 passed, 43 skipped.
- Real built MCP executable and OpenAI request: detached start returned in 21 ms; a zero-timeout waiter timed out; cancellation rejected another waiter while the worker remained running; closing and reconnecting the MCP client preserved the same worker/session; the final wait returned completed output with the requested marker.
- Real filesystem persistence test passes. Filesystem notifications were unavailable in this Mac's private temp directory; the polling fallback correctly observed completion.
- Independent autoreview through P2 is clean on the final candidate.

The live lifecycle proof used the API engine. Signed-in browser/CDP release smoke remains a separate outstanding gate.

PR #459, opened 2026-09-07
chore(mcp): migrate to SDK v2 with legacy client compatibility

### LINKED ISSUE #360: Upgrade `oracle-mcp` to MCP v2 and the modern protocol
### Summary

Upgrade Oracle’s MCP integration to the TypeScript SDK v2 and the modern MCP protocol.

This would replace Oracle’s older MCP foundation with the current server abstractions, simplify several parts of the existing implementation, strengthen schema and transport behavior, and make newer MCP capabilities available for future Oracle features.

Compatibility with clients using the previous MCP protocol should be maintained.

### Why Oracle should upgrade

Oracle currently uses the monolithic v1 MCP SDK and manually manages parts of the stdio server lifecycle and tool-handler plumbing that are handled more directly by SDK v2.

The upgrade would be useful even if Oracle added no new tools or user-facing behavior. It would reduce bespoke protocol code, improve validation and transport handling, and provide a better foundation for features that are awkward to implement against the existing integration.

Immediate improvements include:

- **Current MCP protocol support**  
  Modern MCP clients could negotiate and use the current protocol rather than Oracle remaining limited to the previous protocol generation.

- **Simpler server lifecycle management**  
  SDK v2 provides supported abstractions for stdio startup, protocol negotiation, connection handling, shutdown, and transport errors. Oracle would no longer need to reproduce as much of that lifecycle itself.

- **Smaller runtime dependency surface**  
  The v2 SDK separates server and client packages. Oracle’s runtime needs only the server package, while the client package can remain limited to tests and development tooling.

- **Cleaner schema definitions**  
  Oracle’s MCP tools currently use raw Zod-shape patterns partly to accommodate the older SDK. V2 supports explicit Standard Schema objects, allowing tool input and output schemas to be defined and reused directly.

- **Stronger structured-output validation**  
  Existing `structuredContent` responses can be validated against their declared output schemas before crossing the protocol boundary, making malformed tool results easier to catch at their source.

- **Less low-level server plumbing in tool handlers**  
  Request-scoped context can provide logging, notifications, cancellation signals, and client information without passing internal server objects into tool implementations or reaching through private server properties.

- **More robust stdio behavior**  
  The modern transport improves input buffering, closed-stream handling, shutdown behavior, and transport error handling.

- **Clearer separation of server definition and transport**  
  Oracle’s tools and resources can be registered through a reusable server factory rather than being coupled directly to one transport startup path.

### Existing Oracle behavior to preserve

The upgrade should preserve Oracle’s current MCP interface and execution behavior:

- `consult`
- `chatgpt_image`
- `project_sources`
- `sessions`
- `oracle-session://{id}/{kind}` resources

Tool names, descriptions, parameters, defaults, structured outputs, resource identifiers, dry-run behavior, and session behavior should remain equivalent after the migration.

### Further improvements enabled by MCP v2

The migration would also make the following focused improvements possible in later work:

- **Semantic progress events**  
  Report stages such as launching the browser, uploading attachments, selecting the model, waiting for the response, and downloading artifacts.

- **Request cancellation**  
  Propagate the request-scoped abort signal into foreground browser and provider operations.

- **Resource subscriptions**  
  Allow compatible clients to subscribe to session metadata, log, result, or artifact changes instead of repeatedly polling resources.

- **In-band confirmations**  
  Represent approval-required operations as protocol interactions rather than special Boolean retry fields understood only by Oracle-aware callers.

- **Interactive authentication flows**  
  Support resumable login, OTP, and manual-action-required states through structured protocol interactions.

- **Cacheable tool discovery**  
  Allow compatible clients to cache Oracle’s largely static tool and resource catalog.

- **Deterministic discovery responses**  
  Keep tool and resource definitions stable across connections, improving reproducibility and potential host-side prompt caching.

- **Standard asynchronous tasks**  
  Evaluate the MCP Tasks extension for long-running consultations, standardized task status, cancellation, and result retrieval.

- **Durable task handles**  
  Associate MCP task identifiers with Oracle sessions so results can be retrieved after the original request connection ends.

- **Completion subscriptions**  
  Notify compatible clients when long-running tasks or session resources change state.

- **Cleaner separation of logs and progress**  
  Keep detailed diagnostics in Oracle logs or stderr while exposing concise, structured progress through MCP.

- **Modern HTTP transport**  
  Reuse the same MCP server definition for a future HTTP transport without duplicating Oracle’s tool-registration logic.

- **Stateless hosted deployment**  
  Support future remote Oracle deployments where independent MCP requests can be routed across service instances.

- **Per-request capability handling**  
  Adapt behavior using the identity and capabilities supplied with each request.

- **Gateway-level routing and authorization**  
  Allow future HTTP deployments to route or authorize specific MCP methods using standard protocol metadata.

steipete:
Fixes #360.

Migrate the existing MCP server to SDK v2 while preserving legacy clients, tool/resource contracts, both executable entrypoints, and the detached consultations/waits now landed in #430. The wait handler reads the SDK v2 request signal; consultation progress uses request-scoped logging. A caller timeout, cancellation, or transport shutdown never cancels the detached worker.

The compatibility matrix covers SDK v1, SDK v2 legacy, and SDK v2 pinned to 2026-07-28, each through standalone oracle-mcp and the oracle oracle-mcp alias. The built worker matrix checks wait timeout, explicit cancellation, an interrupted waiter, reconnect, durable completion, and exactly one provider request. It runs automatically in the suite against a local API fixture and was also run against real OpenAI through a delayed-response proxy.

Validation: 2,064 tests passed, 43 skipped; typecheck/lint/build/docs passed. All six real API lifecycle scenarios passed, returning the detached session within 33 ms. Final review and CI links are in the proof comment. Signed-in browser execution is outside this API lifecycle proof.

This branch is rebased onto main with #458, #430, #406, #460, #461, and #383 already landed. Changelog entries remain in #462, which must land afterward. No merge or release is performed here.

### DIFF
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -3,12 +3,6 @@ import "dotenv/config";
 import { fileURLToPath } from "node:url";
 import { Command, Option } from "commander";
 import type { OptionValues } from "commander";
-// Allow `npx @steipete/oracle oracle-mcp` to resolve the MCP server even though npx runs the default binary.
-if (process.argv[2] === "oracle-mcp") {
-  const { startMcpServer } = await import("../src/mcp/server.js");
-  await startMcpServer();
-  process.exit(0);
-}
 import { resolveEngine, type EngineMode, defaultWaitPreference } from "../src/cli/engine.js";
 import { shouldRequirePrompt } from "../src/cli/promptRequirement.js";
 import { resolveDashPrompt } from "../src/cli/stdin.js";
@@ -2967,6 +2961,12 @@ program.action(async function (this: Command) {
 });
 
 async function main(): Promise<void> {
+  // npx runs the default binary; let the MCP transport own the alias lifetime.
+  if (process.argv[2] === "oracle-mcp") {
+    const { startMcpServer } = await import("../src/mcp/server.js");
+    await startMcpServer();
+    return;
+  }
   if (perfTraceArgs.error) {
     console.error(`error: ${perfTraceArgs.error}`);
     console.error("(use --help for usage)");
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -132,3 +132,14 @@ The MCP response includes `structuredContent.images[]` with the saved file path,
   - Claude Code with local macOS Chrome: `oracle bridge claude-config --local-browser > .mcp.json`
 - Tools and resources operate on the same session store as `oracle status|session`.
 - Defaults (model/engine/etc.) come from the effective Oracle CLI config; see `docs/configuration.md`, `~/.oracle/config.json`, and project `.oracle/config.json` files.
+
+### Lifecycle compatibility check
+
+After building, run `node scripts/mcp-lifecycle-proof.mjs` to exercise SDK v1,
+SDK v2 legacy, and SDK v2 modern clients through both executable entrypoints.
+It starts real detached CLI workers against a local API fixture and checks caller
+timeouts, request cancellation, transport reconnects, durable completion, and a
+single provider submission. The standard test suite also runs this matrix.
+For a real OpenAI run, add `--live-key-file <private-key-file>`; the harness uses
+an authenticated upstream request and delays its reply until the lifecycle
+assertions finish. This does not exercise signed-in browser execution.
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -2,6 +2,8 @@
 
 `oracle-mcp` is a minimal MCP stdio server that mirrors the Oracle CLI. It shares session storage with the CLI (`~/.oracle/sessions` or `ORACLE_HOME_DIR`) so you can mix and match: run with the CLI, inspect or re-run via MCP, or vice versa.
 
+The server uses MCP SDK v2 and serves both modern and legacy protocol clients over stdio. Existing tool inputs, defaults, structured outputs, and `oracle-session://` resources are preserved. The SDK owns protocol negotiation and transport shutdown; the server definition is shared between protocol generations. HTTP transport, authentication, subscriptions, and protocol task APIs remain separate from this migration.
+
 ## Let Them Fight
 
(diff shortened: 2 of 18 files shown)


steipete:
Final maintainer proof on `7e6cc47197eef8f2bba0dc36ab5f5bd05f93a339`: all four CI jobs passed at https://github.com/steipete/oracle/actions/runs/34113762454.

The refreshed review exposed a real regression in the documented `oracle oracle-mcp` alias: startup returned and the CLI called process.exit before serving a request. The three alias cases failed against the previous built head while all standalone cases passed. Dispatch now returns from the normal CLI main function and leaves lifetime to the MCP transport, without falling through to argument parsing.

The real built binaries now pass six client/entrypoint combinations: SDK v1, SDK v2 legacy, and SDK v2 pinned to 2026-07-28, each through standalone `oracle-mcp` and the CLI alias. Each combination performs tool discovery, a consultation preview, durable resource reads, and shutdown. The full suite passed 1,998 tests (43 skipped), with typecheck/lint/build and final branch autoreview through P2 clean. The earlier real OpenAI consultations proved all three protocol modes; the alias repair changes no provider execution code.

Integration gate remains: #430 and this branch were independently prepared from main. After #430 lands, adapt its detached/wait imports, cancellation signal, and request-scoped logger to SDK v2 and re-prove the combined tree before this PR lands. Changelog credit remains in #462.

steipete:
The #430 + #459 integration gate passed on `a457b03bd80495c2aa78994a9b69a1d50239b1c8`, rebased onto main `aac815aef64ac574ca0071565ae29a797ebff59a`.

The landed detached launcher is preserved, consultation logs use the SDK v2 request context, and the wait handler observes `context.mcpReq.signal`. Both executables were exercised with SDK v1, SDK v2 legacy, and SDK v2 pinned to 2026-07-28.

All six combinations passed against a local API fixture and real OpenAI through a delayed-response proxy: detached return, wait timeout, explicit cancellation, transport shutdown with a pending waiter, reconnect, durable completion, and exactly one provider request. Real API detached returns were at most 33 ms in this run. The proxy delays replies so the lifecycle assertions are deterministic; this does not claim signed-in browser proof.

Validation: 2,064 tests passed, 43 skipped; typecheck/lint/build/docs passed. Local and final branch autoreview through P2 are scoped-clean. All four CI jobs passed on the exact head: https://github.com/steipete/oracle/actions/runs/34116089978.

The former integration blocker is resolved. #462 is stacked on this candidate and retains the final notes/dependency delta. No merge or release was performed.

PR #461, opened 2026-09-07
fix(browser): preserve and report harvest identity conflicts

### LINKED ISSUE #442: Post-harvest metadata can identify one conversation while the persisted transcript and answer belong to another
First, a correction to my own comment on #390. I described this 2026-08-26 case there as a metadata-only problem. It is not. Rechecking the saved artifacts, the run also persisted the other run's answer.

## The inconsistency

Two browser-mode runs, 2026-08-26, on one persistent Chrome profile. C is `pstack-adjud-verdict-1acc17`, D is `pstack-adjud-matt-4896b9`. They shared Chrome pid 93388 on port 55738 and held separate slot leases (`009c65a3`, `e87f2514`) and separate controller pids (93053, 4123).

D's current record identifies two different conversations at once:

| where | value | conversation |
|---|---|---|
| `browser.runtime.conversationId` | `6a8e8318-...` | D |
| `browser.harvest.conversationId` | `6a8e8318-...` | D |
| `browser.archive.conversationUrl` | `chatgpt.com/c/6a8e822c-...` | C |
| `artifacts[0].sourceUrl` | `chatgpt.com/c/6a8e822c-...` | C |
| `artifacts/transcript.md` header | `Conversation: .../6a8e822c-...` | C |
| first line of the stored answer | `[run: pstack-adjud-verdict-1acc17]` | C |

A check limited to the runtime and harvest identity fields sees a consistent, correctly identified run. The stored answer is a different run's.

Attribution here does not need the answer bodies. Every prompt I send opens with an instruction to emit `[run: <slug>]` as the first answer line, as a cross-wire tripwire. So:

- D's prompt asked for `[run: pstack-adjud-matt-4896b9]`.
- The answer stored under D, in both `output.log` and `transcript.md`, opens `[run: pstack-adjud-verdict-1acc17]`.
- The first line of `browser.harvest.lastAssistantSnippet` in D's current record is `[run: pstack-adjud-matt-4896b9]`.

D asked for D, the stored answer is C's, and D's later harvest saw D's. All four marker lines are in the gist.

C's own session directory holds no transcript, no artifacts entry, and no answer in its logs. Grepping every session directory under `~/.oracle/sessions` for C's marker returns only D's `transcript.md` and D's `output.log`.

## Timestamped events

| time (UTC) | source | event |
|---|---|---|
| 06:05:09.813 | C `createdAt` | C's session record created |
| 06:09:04.386 | D `createdAt` | D's session record created, while C was still running |
| 06:13:10.409 | C `harvest.harvestedAt` | a harvest recorded on C, `state: running` |
| 06:38:05 | mtime of D's `output.log` and `transcript.md` | D's artifacts written |
| 06:38:05.402 | D `completedAt` | D records completion |
| 06:39:13.225 | D `harvest.harvestedAt`, and the mtime of D's `meta.json` | D's harvest recorded |

The artifact mtimes matter: `transcript.md` and `output.log` were last written at 06:38:05 and the metadata file at 06:39:13, so whatever produced the harvest record did not rewrite the stored answer.

The Chrome profile's own History database timestamps navigations to both conversations. The profile was shared by both runs, so a row does not say which run navigated. With that caveat:

```
06:05:33  .../6a8e822c   (C)
06:09:30  .../6a8e8318   (D)
06:09:44  .../6a8e822c   (C)
06:09:50  .../6a8e8318   (D)
06:09:51  .../6a8e822c   (C)
06:27:26  .../6a8e8318   (D)
06:38:00  .../6a8e822c   (C)
06:39:09  .../6a8e8318   (D)
06:39:14  .../6a8e822c   (C)
```

D's own conversation exists and was navigated to 26 seconds after D was created. The two conversations then alternate three times inside twenty seconds. The last navigation before D's completion is to C's conversation, and the last before D's harvest record is to D's.

## The retained copy, and what it does not establish

I kept a copy of D's `meta.json` whose content has no `browser.harvest` block and whose `browser.runtime` carries `chromeTargetId 5C2F2C4C...` and `conversationId 6a8e822c-...`, C's identifiers, alongside D's own `controllerPid 4123`. The complete diff against the current file is one added harvest block and three changed scalars, published in the gist.

That copy's own file was written 6 minutes 38 seconds after the current `meta.json`. So it is not a copy taken from the live path at that moment, and I cannot establish when its bytes were read or by what command. I searched shell history for it and found nothing. It is corroboration that an earlier serialization carried C's identity in D's runtime block. It is not proof that any particular operation changed those fields, and I am not claiming a mechanism from it.

Two observations I offer as candidates rather than diagnoses. C's current record holds runtime target `34EDA741...` and harvest target `5C2F2C4C...`, two targets for one conversation. `5C2F2C4C...` is also the value in the retained copy. And C's harvest was recorded while C was still running and another run was live on the same profile.

## Why this is separate from #390

#390 is the capture taking another run's answer, which also happened here. What I think is distinct enough for its own thread is the end state: identity metadata that agrees with itself while the persisted artifact points elsewhere, with nothing in the CLI surfacing the disagreement. Forensics limited to the runtime and harvest fields would miss it, so matching values there are not enough to establish a correct capture.

## What would help

- When a harvest resolves an identity that differs from the stored runtime or artifact identity, surface the disagreement and keep both values rather than leaving only one.
- After a harvest, compare the resolved conversation against the identity already recorded in the artifacts and the transcript. If they disagree, mark the session inconsistent. That check needs no new data.
- In verbose or debug output, record timestamped target and conversation ids at bind, at capture, and at harvest. I asked for this on #390 too, and it is what would have made this reconstructible rather than inferred.

## Environment and limits

`oracle --version` reports 0.17.3. The Homebrew Cellar holds only 0.17.3, with an install receipt dated 2026-08-14T00:19:21Z. Session metadata carries no version field, so that is machine state rather than a per-session record. Browser engine, `--browser-manual-login`, persistent profile, GPT-5.6 Sol at Pro, `maxConcurrentTabs` 3, `archiveConversations: never`, macOS.

I do not have a deterministic reproduction, and I have not tested current main. I do not know which command produced C's harvest record. 0.17.3's `output.log` has no timestamps and carries no target or conversation ids, and `meta.json` stores current values rather than a history, so what I have is an end state plus the browser's own navigation times, not an execution trace. I still have the profile and both session directories and can run an instrumented build against them.

## Evidence

https://gist.github.com/postoso/049d81263d37a580f4eb119e95574fcf

Redacted `meta.json` for both runs, the retained copy, the complete diff between them, the operational portion of both logs, the four run-marker lines, the Chrome history rows, file times, and the commands behind each value. Prompts and answer bodies are omitted apart from the marker lines themselves.

steipete:
Fixes #442.

A later harvest can agree with current runtime metadata while the saved transcript belongs to another conversation. Compare the latest stored runtime/archive/artifact identities and bounded transcript headers with the observed harvest. Retain both identities under `browser.harvest.integrity` and preserve the original files. An implicit mismatch now fails before exporting the new answer; an explicit `--browser-tab` override remains usable but records the conflict and shows a warning.

Recorded provenance that cannot be checked is `unverified`. Download URLs are not mistaken for conversation URLs, and transcript reads stay inside the session's artifact directory. Contradictory saved identities are checked even without a new conversation ID, and repeated unidentified observations cannot clear a previously established mismatch. Existing session status values are preserved, and the normal session viewer displays the durable warning.

Validation: 2,004 tests passed, 43 skipped; typecheck/lint/build/docs checks pass. A real-filesystem run of the built persistence owner reproduced the issue where metadata agrees but the transcript header differs, retained the original transcript/log, and made the warning visible through the built CLI. Explicit-target inspection remained allowed. Independent autoreview through P2 is clean on the final candidate.

This detects the inconsistent stored result; it does not establish or fix the first cross-session ownership divergence reported in #390. Changelog credit is collected in the final notes PR. Thanks @postoso for the paired forensic evidence.

### DIFF
--- a/docs/sessions.md
+++ b/docs/sessions.md
@@ -136,6 +136,20 @@ oracle status --clear --hours 168   # delete sessions older than a week
 
 Every run gets a default slug derived from the prompt. Override with `--slug "my-thing"` for stable names you can reference later (`oracle session my-thing`).
 
+## Browser harvest identity
+
+Browser harvest and live-tail compare the observed conversation with saved
+runtime, archive, artifact-source, and transcript-header identities. A mismatch
+is retained under `browser.harvest.integrity` and shown as a browser warning;
+an implicit harvest fails with `conversation-identity-mismatch` before exporting
+the newly harvested answer. Existing transcripts and answer logs are preserved.
+
+An explicit `--browser-tab` override still permits inspecting another target,
+but records the mismatch and does not reassign the original capture. Unavailable
+recorded transcript headers or unreadable recorded conversation URLs are marked
+`unverified`. Matching known conversation IDs
+does not by itself prove that an answer belongs to the original prompt.
+
 ## Naming conventions
 
 Pair `--slug` with conventional prefixes for browseability:
--- a/src/cli/browserTabs.ts
+++ b/src/cli/browserTabs.ts
@@ -7,7 +7,6 @@ import {
   collectChatGptTabs,
   DEFAULT_REMOTE_CHROME_HOST,
   DEFAULT_REMOTE_CHROME_PORT,
-  extractConversationIdFromUrl,
   formatBrowserTabState,
   harvestChatGptTab,
   sessionMatchesTab,
@@ -18,6 +17,8 @@ import {
   recoverConversationTab,
 } from "../browser/recoverConversation.js";
 import { resolveOutputPath } from "./writeOutputPath.js";
+import { persistBrowserHarvest } from "./harvestIntegrity.js";
+import type { BrowserHarvestIntegrity } from "../sessionManager.js";
 
 const LIVE_POLL_MS = 2000;
 const DEFAULT_STALL_THRESHOLD_MS = 60_000;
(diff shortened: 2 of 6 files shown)


steipete:
Final maintainer proof on `b6beffde46405441859dfcec9f4cf5d3ed737d56`: all four CI jobs passed at https://github.com/steipete/oracle/actions/runs/34113668144; final branch autoreview through P2 is scoped-clean.

The repeated-harvest review finding was reproduced: a later observation without a conversation ID removed the warning despite saved contradictions. The repair compares saved identities independently and preserves a prior mismatch when a new observation cannot verify it. Explicit inspection remains available; implicit known conflicts still fail before export.

The full suite passed 2,004 tests (43 skipped), with typecheck/lint/build passing. The compiled persistence owner ran against real session files, followed by the built CLI session viewer: repeated unidentified observations retained the warning and last known harvest ID, original transcript/log bytes and top-level status stayed unchanged, and explicit override remained recorded. Focused regressions also cover a first unidentified observation of contradictory saved records.

Fixes #442. This does not establish the first ownership divergence required for #390 and does not close that causal investigation. Changelog credit remains in #462.

PR #469, opened 2026-09-08
fix(browser): retire owned tabs after recovery

### LINKED ISSUE #435: Browser mode: recovered/error runs can release the lease while retaining Oracle-owned tabs indefinitely

steipete:
Recoverable browser failures can release their lease while leaving an Oracle-created tab indefinitely, even after a successful harvest or reattach. Browser-level CDP connections could also keep an incomplete controller alive after its page connection was detached.

Record explicit target ownership and save the full recovered answer plus completed session metadata before retiring the captured target. Per-tab claims survive document replacement and reserve retirement against a new controller; current session/lease ownership and generation checks preserve borrowed, kept, legacy, reclaimed, or active targets. Connection cleanup now releases the browser transport independently of whether its tab must remain available for recovery.

There is no automatic retention TTL. An interrupted retirement reservation remains conservative: the saved answer is available, and a new run should use another tab. Host ownership details remain excluded by the existing bridge result allowlist.

Fixes #435. Thanks @lhysin for the detailed lifecycle report.

Validation:
- Full suite: 2,258 passed / 45 skipped; build, formatting, lint/typecheck passed.
- Built main CLI before/after proof against real isolated Chrome: main retained the recovered owned tab and left capture status in error; the candidate saved the complete answer, marked completion, and retired only that tab.
- The real-Chrome matrix also passed borrowed/kept/active-generation/active-controller/reclaimed preservation and refusal of a new controller after retirement reservation. It includes a real same-origin page reload.
- Signed-in built CLI Pro run: deliberate timeout retained its target and the controller exited naturally; later reattach saved the full answer and retired only its owned tab. Peer tabs and Chrome survived.
- Regressions cover failed answer/model/session writes, optional identity-capture failure, host aliases, other sessions and shared leases, current generation controls, reservation ownership, and transport cleanup with target preservation.
- Local and committed-branch P0–P2 autoreview are clean. CI passed on `3f5529fb195fd7605d2d67fb0568cef9d8896bd6`: https://github.com/steipete/oracle/actions/runs/34178930225.

The changelog entry is deferred to the final notes PR so sibling implementations do not conflict.


steipete:
Final-head proof for `3f5529fb195fd7605d2d67fb0568cef9d8896bd6`:

- Full suite: 2,258 passed / 45 skipped; build, formatting, lint/typecheck passed.
- `node scripts/recovery-retirement-proof.mjs --baseline-cli <built-main-cli>` reproduced retained owned-tab/error status on main. The candidate matrix saves the complete answer and completes the session before retiring the owned tab; borrowed, kept, generating, active-controller, reclaimed, and peer targets survive. A same-origin reload preserves ownership and a retirement reservation refuses a new controller.
- Signed-in built CLI Pro run: a deliberate timeout left the target recoverable while the controller exited naturally. Later reattach saved the complete answer containing `ORACLE_PERSISTENT_RECOVERY_OK` and reported “Retired Oracle-owned browser tab after saving the recovered answer.” Chrome and other tabs survived.
- Local and final branch P0–P2 autoreview: scoped-clean.
- Exact-head CI: all four jobs passed, including Linux browser proofs and Windows shared-Chrome lifecycle: https://github.com/steipete/oracle/actions/runs/34178930225.
- The combined six-PR tree also passed the real-Chrome retirement matrix and a signed-in Latest/Pro Web Search with bundled attachments.

Fixes #435. Changelog and thanks @lhysin are collected in #470, to merge last.

PR #473, opened 2026-09-09
feat(browser): verify persisted ChatGPT Scheduled tasks

derekszen:
## Summary

Supersedes #322 by rebuilding the Scheduled-task browser workflow on current `main` and adding a fail-closed persistence check.

- Add `--browser-scheduled-task` and `--browser-pin-conversation` to the current CLI/config path.
- Keep Scheduled submissions on `/scheduled`, skip model switching, and reject incompatible follow-up/research/archive modes.
- Allow ChatGPT's Scheduled confirmation UI to settle, then navigate to the Scheduled page and require a matching active task row before reporting success.
- Verify the optional conversation pin through the observed `Unpin` state.

The old #322 false-success path accepted stable confirmation text without proving that ChatGPT persisted a task. This implementation does not treat assistant prose as task proof.

## Validation

- Focused browser/CLI tests: 349 passed.
- Scheduled-task verification tests included for matching active rows and inactive/missing rows.
- `pnpm exec tsc -p tsconfig.build.json --noEmit` passed.
- `pnpm run check` passed (format, typecheck, lint).

A live authenticated ChatGPT Scheduled-task smoke is still required before merge to confirm the current task-list DOM selectors and provide redacted proof. The check fails closed if the row cannot be found.


steipete:
Closing this revision because the persistence check can report success for an unrelated existing task. `verifyScheduledTask` accepts three substring hits among the first six long prompt tokens, including repeated tokens; it does not bind the result to the task created by this submission or verify its schedule. For example, `Create a monthly task titled CHECK_TASK_OK` accepts a row containing `Monthly task CHECK_OTHER_OK`.

That does not meet the durable-result requirement from #322. Scheduled-task support needs a task-specific identity/schedule contract and authenticated persistence proof before adoption. I have retained that decision in the maintenance report; no contributor revision is requested.

PR #474, opened 2026-09-10
fix(cli): refuse stale browser harvest output

pdurlej:
Browser harvest could persist a stale assistant turn while a conversation was still hydrating. New ChatGPT browser sessions now record the rendered, committed user turn and require a matching turn plus a paired assistant answer before implicit harvest. A mismatch preserves the existing metadata and output.

Fingerprints bind the rendered text to its stable provider message ID, preserve case and meaningful whitespace, and update after each submitted follow-up. This prevents an earlier identical “Continue” prompt from passing during incremental hydration and survives conversation remounting. Capture compares stable message IDs observed before and after submission, without carrying DOM offsets across that boundary. A pending marker survives startup and preparation failures, so interrupted new sessions cannot be misclassified as legacy sessions. Capture and harvest wait within the saved input timeout, following the maintainer direction in #409.

Older saved sessions never recorded the assembled file context needed for exact submitted-turn verification. They retain pairing-only recovery with an explicit unverified warning. Explicit alternate-tab inspection retains its existing behavior and still requires a paired answer. Thanks @pdurlej.

Validation includes the full test suite, regression coverage for startup failures, delayed rendering, execution-context replacement, and follow-up capture, plus the actual built CLI against real isolated Chrome. The live checks exercise delayed hydration beyond five seconds, unrelated/prefix/suffix rejection with byte-for-byte output preservation, explicit alternate-tab inspection, follow-ups, system/file context, unconfirmed submissions, remounted history, and legacy Markdown/list/ZIP recovery. A separate three-turn built CLI run verifies saved hashes from rendered user turns while earlier DOM turns are transiently unmounted. A reproduced TUI-test timing flake now waits for the expected async startup call instead of assuming it happens within 100 ms. Required branch autoreview and exact-head CI are recorded in the proof comment.

This branch also carries the exact controller-identity/Windows verification repair from #479 after its standalone Windows CI reproduced that known failure. Land #479 first; the duplicate patch becomes a no-op. All changelog notes are centralized in #481, which must land last.


steipete:
Verified head `3a866b7555a865a29d4ef1fd44a96ef1942e4106`.

Fresh CI: https://github.com/steipete/oracle/actions/runs/34657177330 — CI BLOCKED as of 23:35 UTC. Ubuntu, Windows, and Linux browser proofs are successful. macOS still has no runner after about 23 minutes and one job-only retry. This PR must wait for that job to pass before merging. Required branch autoreview through P2 is scoped-clean.

Local verification: all 110 focused session/harvest/fingerprint regressions, typecheck, lint, and build passed. The final full local suite had 2,366 passes and the already-diagnosed host-profile discovery timeout in the service fixture; #481 makes that fixture hermetic. An independently reproduced TUI startup test flake is repaired here: the same assertion now waits for async startup within five seconds, and always restores argv/TTY. Its tests and isolated review passed.

Actual built CLI + isolated real Chrome: all 15 harvest scenarios passed, including hydration beyond five seconds, an earlier identical “Continue” turn, and recovery after earlier DOM turns are unmounted, unrelated/prefix/suffix rejection with original metadata and output preserved, explicit alternate-tab inspection, follow-ups, system/file context, unconfirmed submission refusal, and legacy Markdown/list/ZIP recovery with an unverified warning. A separate three-turn real CLI/Chrome run with transient virtualization after native commit detection verified saved fingerprints against delayed, rendered user turns and their stable message IDs, including a Markdown follow-up. These use synthetic provider pages, not a live provider answer.

The preparation-failure regression proves that both a new pending session and a rerun with an old fingerprint retain a null pending marker while discarding stale browser endpoints and model-selection evidence. Older tagged sessions intentionally keep pairing-only recovery: their metadata cannot prove the exact submitted file context.

The earlier Windows failure in run 34651774733 was the same controller-identity/lease timing problem fixed by #479. This branch carries that exact patch; land #479 first. Changelog notes and the fixture isolation repair remain centralized in #481, which lands last. Contributor credit is preserved for @pdurlej.

The accepted GitHub review finding is repaired: the durable hash uses `data-message-id`, not the mounted DOM offset. The real CLI remount proof first captures a fingerprint with earlier turns mounted, removes those earlier turns, then recovers the same answer successfully. An older identical prompt with a different message ID is still rejected.

Capture also compares pre-submit and post-submit stable message IDs; it no longer reuses a pre-submit DOM-array boundary. A regression keeps the mounted-array length unchanged while replacing an old turn with the new repeated prompt, and capture correctly records the new message.

CI run 34656146491 also exposed synthetic proof pages without user-message IDs, causing the service legacy-completion timeout. Those fixtures now emit stable IDs like the provider DOM; assertions and timeout bounds are unchanged.

The repaired real Chrome service admission/cancellation proof passes locally, including the legacy protocol path, queueing, cancellation guards, shared-browser preservation, and cleanup.

The macOS job waited about 19 minutes without a runner; only that unstarted job was requeued. The successful Ubuntu, Windows, and Linux browser-proof results remain on the same head.

PR #476, opened 2026-09-10
fix(cli): cancel detached browser sessions truthfully

pdurlej:
Long browser consultations need a detached worker even while the foreground CLI waits for them. Ctrl-C now requests cancellation from that worker, and the worker persists cancelled session/model state while cleaning up resources it owns. Cancellation and completion races preserve the terminal outcome, including cancellation during reattachment and output persistence.

The maintainer repair also preserves explicitly kept tabs during final target retirement and removes a late cancellation marker when worker completion wins. API wait/detach behavior remains governed by the existing separate policy. Thanks @pdurlej.

Validation: 111 focused cancellation/session tests plus target-retirement and detach coverage pass; lint/typecheck and build pass. The real built CLI/Chrome matrix verifies normal cancellation (exit 130), keep-browser cancellation (exit 130 with the target retained), and a paused foreground process receiving Ctrl-C after worker completion (exit 0). All three verify worker exit, terminal model/session metadata, and marker removal. Required isolated review and exact-head CI are recorded in the proof comment.

The changelog entry is centralized in the final notes/dependency PR for this batch.

### DIFF
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -45,8 +45,18 @@ import {
 import { copyToClipboard } from "../src/cli/clipboard.js";
 import { isGpt6ProAlias } from "../src/cli/browserConfig.js";
 import { buildMarkdownBundle } from "../src/cli/markdownBundle.js";
-import { shouldDetachSession, stopDetachedWorker } from "../src/cli/detach.js";
-import { launchDetachedSession } from "../src/cli/detachedSession.js";
+import {
+  shouldDetachSession,
+  shouldExitAfterTopLevelSigint,
+  stopDetachedWorker,
+} from "../src/cli/detach.js";
+import {
+  clearDetachedSessionCancellation,
+  detachedSessionCancellationPath,
+  launchDetachedSession,
+  requestDetachedSessionCancellation,
+  waitForDetachedSessionCancellation,
+} from "../src/cli/detachedSession.js";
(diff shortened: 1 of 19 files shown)


steipete:
Verified head `6ed19a1126cea7db3b66db7475b759bd7d94fc90`.

Fresh CI: https://github.com/steipete/oracle/actions/runs/34650063969 — SUCCESS across Ubuntu, macOS, Windows, and Linux browser proofs. Required branch autoreview through P2 is scoped-clean after the maintainer repairs. Local cancellation/session tests (111), target-retirement/detach regressions, typecheck/lint, and build pass.

Real built CLI + Chrome proof:
- Ordinary detached-browser cancellation: exit 130, session/model cancelled, worker exits, cancellation marker removed, owned target closes.
- `--browser-keep-browser`: exit 130 with the exact target retained, worker exited, and no marker left.
- Completion wins: pause the foreground parent after submission, let the worker complete and exit, then deliver Ctrl-C and resume the parent. Exit remains 0, session/model remain completed, and the late marker is removed.

The paused-parent case accounted for the exited child remaining a zombie until its parent could reap it; it did not treat a live worker as complete. Each case submitted one synthetic prompt and used an isolated profile.

Ready for squash after maintainer authorization. Preserve Piotr Durlej's credit; the changelog bullet is in #481.

PR #479, opened 2026-09-11
fix(browser): retry unavailable controller identity probes

steipete:
Windows CI intermittently exhausts the tab-lease tests' five-second budget while running real PowerShell identity probes, whose individual timeout is also five seconds. A failed self-identity probe was additionally cached for the controller lifetime, preventing later lock checks from recovering usable identity evidence.

Retry unavailable self identities on the next lookup, retain shared in-flight probes and successful caching, and keep peer identities uncached. Give the native Windows registry tests a 30-second budget for multiple bounded probes and cleanup; no tests or ownership checks are removed.

Validation: new regressions failed twice before the fix; focused identity/profile/registry tests pass (50 passed). Typecheck, lint, build, and isolated Codex review through P2 pass. Full-suite and real two-controller Chrome lifecycle proof are running; exact-head CI will be recorded in a follow-up proof comment.

The user-facing changelog entry will be included in the final notes/dependency PR for this batch.

### DIFF
--- a/docs/windows-work.md
+++ b/docs/windows-work.md
@@ -14,6 +14,8 @@ Read this file whenever you're working from Windows and add new findings so the 
 
 Future Windows gotchas belong here. Update this doc when you learn something new.
 
+- Tab-lease tests run real PowerShell process-identity probes (up to five seconds each). Their Windows test budget must cover multiple probes and registry cleanup. Failed self-identity probes are retried on the next lookup; only a successful identity is cached for the controller lifetime.
+
 - A fresh Windows worktree with `core.autocrlf=true` can make `oxfmt --check` flag otherwise unchanged files. Use LF checkout contents for validation and inspect the staged diff to keep checkout-only line-ending changes out of the PR.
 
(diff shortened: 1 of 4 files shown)


steipete:
Verified head `6aa3727a1b4382fbabb27397124c2220ded20b43`.

- Fresh CI: https://github.com/steipete/oracle/actions/runs/34648494860 — SUCCESS across Ubuntu, macOS, Windows, and Linux CDP/browser proofs.
- Local full suite: 181 files passed, 2,333 tests passed (18 files/45 tests retain their existing opt-in/platform skips).
- New identity regressions reproduced two failures before the fix; identity/profile/registry coverage passes after it.
- `pnpm run lint` and `pnpm run build` pass.
- `node scripts/shared-chrome-lifecycle-proof.mjs` passes with two real built CLI controllers and real Chrome: the peer survives owner exit, final registry is empty, Chrome exits, and the endpoint closes.
- Required branch autoreview through P2: scoped-clean, no actionable findings.

Ready for squash after maintainer authorization. This repairs the cause of the Windows default-branch failure in run 34182600456; main itself remains unchanged until this PR lands. The changelog entry will be in the final notes/dependency PR.

PR #485, opened 2026-09-12
fix(browser): warn before implicit default model downgrades

### LINKED ISSUE #375: Browser mode's default model now silently switches accounts off their selected newer model (gpt-5.5-pro default × 0.17.2 picker navigation)
## Setup
oracle 0.17.3 (brew tap), macOS arm64, browser lane with a persistent `--browser-manual-login` profile whose ChatGPT account has **GPT-5.6 Sol** selected with **Pro** effort.

## What changed
Pre-0.17.2, a no-`-m` browser run effectively rode the account's selected model (our runs resolved to GPT-5.6 Pro even though the requested alias was the `gpt-5.5-pro` default). Since 0.17.2's unified-picker navigation, the default alias **actively selects the GPT-5.5 row** — which is now a real, distinct model row in ChatGPT's Advanced → Model menu, not a stale label:

```
Model selection evidence: requestedKey=gpt-5.5-pro; target=GPT-5.5; resolvedLabel=GPT-5.5;
status=switched; strategy=select; verified=yes; source=chatgpt-model-picker
```

The run genuinely executes on GPT-5.5 Pro (ChatGPT's own "Used GPT-5.5 Pro" footer confirms). `verified=yes` faithfully verifies the switch — the switch itself is the surprise.

## Why it bites
Every account sitting on a newer model gets silently downgraded on every no-`-m` browser run. It hit us twice in one evening, including a consequential adjudication run we only caught via the ChatGPT footer afterward. The alias is doing exactly what its name says — the problem is that it's the *default* in a world where the account selection is newer.

## Suggested fixes (any of these would resolve it)
1. **Bump the browser default to the GPT-5.6 family** — our preference: it's the current model and noticeably more thorough. (The `GPT-5.5 Pro/GPT-5.5` branding in `--help` is stale throughout anyway.)
2. Make the no-`-m` browser path default to current-model semantics (keep the account's selected model, apply only the effort control) — the pre-0.17.2 effective behavior.
3. At minimum, warn loudly when a switch away from the account's currently-selected model is caused by the *default* rather than an explicitly requested `-m`.

Workaround we've adopted meanwhile: always pass `-m gpt-5.6-sol --browser-thinking-time pro` (the new fail-closed `pro` level is great, by the way).

Happy to PR whichever direction you prefer.

steipete:
When no model is supplied through the CLI or configuration, Oracle can switch ChatGPT away from an already selected newer model without identifying the default as the cause. Warn before that switch and submission while retaining the established default and model-selection behavior.

Carry implicit-default provenance through CLI, MCP, saved session configuration, and the remote service. Explicit models, configured preferences, and current/ignore strategies remain quiet. A delayed picker probe retains a separate selection deadline so warning detection cannot consume the selection budget.

Fixes #375.

Validation: 45 focused regression tests, typechecking, clean local/branch autoreview through P2, and five built-CLI browser scenarios through one retained connection. All four exact-head CI jobs passed: https://github.com/steipete/oracle/actions/runs/34668020695. The changelog entry is consolidated in the final maintenance-notes PR.

### DIFF
--- a/bin/oracle-cli.ts
+++ b/bin/oracle-cli.ts
@@ -2226,6 +2226,8 @@ async function runRootCommand(options: CliOptions): Promise<void> {
       browserRequestedModel: cliModelArg,
       browserModelLabel: resolveBrowserModelLabel(cliModelArg, activeModel),
     });
+    config.modelIsImplicitDefault =
+      optionUsesDefault("model") && !userConfig.model && !options.browserModelLabel;
     return resolvedOptions.browserResumeConversationUrl
       ? { ...config, resumeConversationUrl: resolvedOptions.browserResumeConversationUrl }
       : config;
--- a/docs/browser-mode.md
+++ b/docs/browser-mode.md
@@ -43,6 +43,13 @@ oracle --engine browser \
 
 You can pass the same payload inline (`--browser-inline-cookies '<json or base64>'`) or via env (`ORACLE_BROWSER_COOKIES_JSON`, `ORACLE_BROWSER_COOKIES_FILE`). Cloudflare cookies (`cf_clearance`, `__cf_bm`, etc.) are only needed when you hit a challenge.
 
+When no model is supplied on the command line or in configuration, Oracle keeps
+its existing browser default. If the visible ChatGPT selection is a newer model,
+Oracle prints a model-selection warning before switching and submitting the
+prompt. Pass `--model` to choose explicitly, or `--browser-model-strategy current`
+to retain ChatGPT's selection. Explicit models, saved model preferences, and
+`current`/`ignore` strategies do not produce this warning.
+
 ## Quick example: attach to your running Chrome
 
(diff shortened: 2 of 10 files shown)


steipete:
Maintainer proof for head `36fb9738f08c7a1710fb8e42d9598d24844926ce`: 45 focused tests and clean local/branch autoreview through P2. The composed batch passed 2,458 tests.

Five actual built CLI clients ran through one retained service/browser connection against synthetic model controls. Implicit and delayed-picker cases warned before switching; explicit, configured, and current-model cases remained quiet. Every case submitted exactly once and captured `ORACLE_WARNING_PROOF_OK`. This tests warning/provenance behavior; the synthetic fixture's effort control is not a signed-in effort proof. A separate signed-in Pro/reattach/harvest control passed in the same connection.

The existing default remains unchanged. All four exact-head CI jobs passed: https://github.com/steipete/oracle/actions/runs/34668020695. The changelog entry is consolidated in #486.

PR #486, opened 2026-09-12
chore: consolidate browser fixes and refresh Zod

steipete:
Consolidate the batch's user-visible browser and remote-image fixes in one ordered Unreleased section, and update Zod from 4.6.0 to the eligible 4.6.1 patch with a consistent lockfile. Node >=24 and the two-day minimum release age remain unchanged; the newer 4.6.2 patch is still inside that age window.

Land this notes PR last, after #487 → #483 → #477 → #471 → #485. Each sibling fix keeps its changelog changes here to avoid squash-merge conflicts. These fixes support patch 0.20.2; this PR does not bump, tag, or publish a version.

Validation: 2,402-test dependency suite; built CLI version, dry-run, and docs checks; clean local autoreview through P2. All four exact-head CI jobs passed (https://github.com/steipete/oracle/actions/runs/34668276052), and the branch review is clean through P2. Hold this PR until the implementation PRs have landed.

PR #487, opened 2026-09-12
fix(browser): reuse the process DevTools connection

### LINKED ISSUE #484: browser: reuse DevTools approval across operations in one process
Chrome's remote-debugging approval applies to each browser WebSocket connection. Oracle currently creates a new connection for target discovery and another for each tab attachment, then closes those connections at the end of the operation. A multi-step run or successive `oracle serve` requests can therefore repeatedly show **Allow remote debugging?** even when Chrome has stayed connected throughout.

The ownership mismatch is in `src/browser/chromeLifecycle.ts`: `listRemoteChromeTargets` and `connectToRemoteChromeTarget` independently call `connectToBrowserWebSocket`, and their cleanup closes the browser transport along with the page session.

Keep one browser WebSocket per endpoint for the lifetime of the Oracle process. Share in-flight connection attempts, use separate flattened page sessions, and detach only the page session during request cleanup. Remove listeners owned by a completed session, preserve existing tab ownership rules, let idle CLI processes exit, and reconnect only after an actual transport disconnect. Separate CLI processes still require separate connections; a long-running `oracle serve` host is the sharing boundary across client commands.

Verification should count WebSocket handshakes during discovery, multiple requests, and recovery using one long-lived host; test concurrent acquisition, cancellation, tab closure, listener cleanup, idle process exit, and genuine disconnect recovery. No browser approval loop is needed to reproduce the source-level ownership issue.

steipete:
Chrome approves each browser WebSocket separately. Oracle previously opened and closed that socket for target discovery and each page operation, so one run or successive service requests could repeatedly show **Allow remote debugging?**.

Retain one browser connection per endpoint for the process lifetime, share pending approvals, and give each page an independently disposable session. Cancellation and approval deadlines release only their own lease; failed handshakes and real disconnects permit reconnection. Page listeners are removed on disposal, command callbacks remain session-bound, failed attachment cleans up only its newly created tab, and idle sockets allow a normal CLI exit.

Fixes #484.

Validation: 51 focused connection/lifecycle tests, a 2,410-test full-suite baseline followed by the focused callback regression, and the composed batch's 2,458-test suite. The compiled transport proof counted one real WebSocket handshake across three discovery/attach/evaluate/dispose cycles and natural child-process exit; repeated real CLI clients also reused one service socket against a synthetic DevTools endpoint. CI now runs the standalone transport proof after an explicit build on every platform.

The instrumented signed-in proof uses one long-lived built service: five model-warning scenarios, a Pro answer, reattach, and harvest have completed with exactly one handshake/upgrade. The same service also delivered the generated PNG with its dimensions and file ID intact. The probe refuses any second handshake. Separate CLI processes still require separate approval; use one long-running service to share approval across client commands. Release notes are consolidated in #486.

Local and committed-branch autoreviews are clean through P2. All four CI jobs passed on final head `465843e5be5b96d3a0b8d6fa32776cf10e651f00`: https://github.com/steipete/oracle/actions/runs/34669035154. Ready for the orchestrator’s squash-merge decision. The Windows run exposed an existing MCP proof cleanup race (EBUSY after signalling a detached worker). The follow-up waits for worker exit before directory removal and adds bounded filesystem retries without dropping any lifecycle assertion.


steipete:
Live proof for head `465843e5be5b96d3a0b8d6fa32776cf10e651f00` used the composed batch build and one long-running built CLI service. Instrumentation covered the WebSocket library's actual HTTP GET upgrade path and refused a second handshake. It recorded one attempt at 2026-09-12T02:44:00.455Z and one upgrade at 02:44:04.376Z, with no subsequent handshake across five CLI warning runs, a signed-in Pro answer, Pro-preserving reattach, matched-identity harvest, generated-image transfer, provider sign-in inspection, and owned-tab cleanup.

The standalone compiled-transport proof independently counted one handshake across three discovery/attach/delayed-evaluation/dispose cycles and natural child-process exit. Real CLI service-routing fixtures also cover repeated requests using one transport. Validation includes 51 focused tests and the 2,458-test composed suite; local and branch autoreview are clean through P2. All four CI jobs passed at final head `465843e5be5b96d3a0b8d6fa32776cf10e651f00`: https://github.com/steipete/oracle/actions/runs/34669035154. The follow-up to the initial Windows CI failure changes only the MCP proof and Windows notes: wait for detached-worker exit before removing its working directory, then use bounded filesystem retries. All six MCP lifecycle cases pass locally with explicit worker-exit and cleanup assertions. The live browser code is unchanged from the instrumented proof.

For transparency, the maintenance session used two Chrome attaches total: one preliminary warning run before correcting its handshake counter, then the single final instrumented run above. There were no per-test reconnects. Release notes are in #486.

PR #497, opened 2026-09-16
fix(browser): force C locale for ps lstart parsing on non-English macOS

zianai:
## Problem

`queryProcessStartTimeMs` in `src/browser/profileState.ts` reads a process's start time via `ps -p <pid> -o lstart=` and parses the output with `Date.parse`. However, `ps` renders `lstart` in the **user's locale**. On a zh_CN macOS system the output looks like:

```
三  9月/16 22:56:52 2026
```

`Date.parse` rejects that string, so the probe returns `null` for **every** pid. Process identity is then permanently unavailable, which silently degrades everything built on top of it: shared-Chrome lease verification, stale-lock recovery, and pid-reuse detection (all the hardening from the recent concurrent-controller work).

On such a machine, three tests fail on current `main` (1866b146):

- `tests/browser/profileState.test.ts > reads the current process start identity` — asserts the probe returns non-null
- `tests/browser/tabLeaseRegistry.test.ts > recovers a lock whose pid was reused by a different process identity` — times out (5s) waiting on null probes
- `tests/browser/tabLeaseRegistry.test.ts > uses legacy lock creation time to detect a reused live pid` — same timeout

## Fix

Force the C locale for the `ps` invocation so `lstart` stays in the parseable `Wed Sep 16 22:56:52 2026` form:

```ts
env: { ...process.env, LC_ALL: "C" },
```

The Windows path is unaffected — it already emits an ISO-8601 string via PowerShell's `ToString('o')`. Any non-English Linux distro benefits the same way.

## Verification

On a zh_CN macOS machine (Node 26.8.2):

- The three tests above go from red (1 assertion failure + 2 timeouts) to green; the two tabLeaseRegistry cases now complete in milliseconds instead of timing out
- Full suite: 2520 passed / 44 skipped / 1 failed — the single remaining failure is `runOracle.request-payload > does not require an Azure deployment for non-OpenAI providers`, caused by a local `ANTHROPIC_BASE_URL` env override on my machine (not present in CI)
- `pnpm run typecheck` and `oxlint` are clean


steipete:
Thanks @zianai! Your locale fix is now landed with regression coverage and credit in #505.

